PR #36, opened 2017-12-06
refactoring Algorithm definition components into separate encryption, kdf, and authentication algorithm suites

mattsb42-aws:
This is an idea I've been playing with to try and make the algorithm suites more comprehensible and less error-prone.

## Status Quo
Each algorithm suite is composed of a large number of individual values that together form the complete algorithm suite specification.

## Problem
These algorithm suite definitions are hard to understand, hard to debug, and easy to mess up.

## Solution
Algorithm suites are not conceptually a grouping of a large number of primitives. They are a grouping of a small number of groupings of related values. Each of these groupings define characteristics of the overall algorithm suite.

I have separated these out into three sub-suites:
1. Encryption : contains only those values that control the encryption behavior
1. KDF : contains only those values that control the key derivation behavior
1. Authentication : contains only those values that control the authentication behavior

Each algorithm suite is then defined as a combination of these three sub-suites.

This lets us move from each algorithm suites having 13 distinct characteristics with no innately relations to:
* Three distinct encryption suites
* Two distinct KDF suites
* Three distinct authentication suites

Each algorithm suite now has four important values: algorithm ID, encryption suite, KDF suite, and authentication suite.

For legacy compatibility, when each algorithm suite is constructed, it still sets all of its expected attributes using the attributes of the sub-suites.

### DIFF
--- a/src/aws_encryption_sdk/identifiers.py
+++ b/src/aws_encryption_sdk/identifiers.py
@@ -173,7 +173,7 @@ class AlgorithmSuite(Enum):  # pylint: disable=too-many-instance-attributes
         self.authentication = authentication
         self.allowed = allowed
 
-        # Encryption Suite Legacy Compatibility
+        # Encryption Values
         self.encryption_algorithm = self.encryption.algorithm
         self.encryption_mode = self.encryption.mode
         self.data_key_len = self.encryption.data_key_length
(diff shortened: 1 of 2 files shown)


src/aws_encryption_sdk/identifiers.py
```diff
@@ -25,116 +25,181 @@
...
-        # All algorithms in this enum are allowed for now.
-        #  This might change in the future.
-        self.allowed = True
+        self.encryption = encryption
+        self.encryption.valid_kdf(kdf)
+        self.kdf = kdf
+        self.authentication = authentication
+        self.allowed = allowed
+
+        # Encryption Suite Legacy Compatibility
```
SalusaSecondus: Please remove "legacy" comment

src/aws_encryption_sdk/identifiers.py
```diff
@@ -25,116 +25,181 @@
...
+        self.encryption_algorithm = self.encryption.algorithm
+        self.encryption_mode = self.encryption.mode
+        self.data_key_len = self.encryption.data_key_length
+        self.iv_len = self.encryption.iv_length
+        self.auth_key_len = self.encryption.auth_key_length
+        self.auth_len = self.tag_len = self.encryption.auth_length
+
+        # KDF Suite Legacy Compatibility
```
SalusaSecondus: Please remove "legacy" comment

SalusaSecondus: Please remove "legacy" comment

mattsb42-aws:
Updated with requested docs/comments changes.

Issue #65, opened 2018-08-15
Update for Pylint 2.0.0


mattsb42-aws:
Pylint 2.0.0 added a bunch of new checks. We need to update to meet them.

Issue #49, opened 2018-04-12
re-enable bandit run in Travis once >1.4.0 is released


mattsb42-aws:
re-enable bandit run in Travis once >1.4.0 is released

The bug that was causing us issues has been fixed[1], but a new version of bandit has not yet been released with these changes.

[1] https://bugs.launchpad.net/bandit/+bug/1749603

Issue #17, opened 2017-09-27
deserialize_header function needs to be refactored


mattsb42-aws:
As pylint and flake8 justifiably point out, `internal.formatting.deserialize.deserialize_header` is too complex. Refactor to simplify.

```
186590df9307:aws-encryption-sdk-python bullocm$ tox -e flake8,pylint
GLOB sdist-make: /Users/bullocm/git/aws-encryption-sdk-python/setup.py
flake8 inst-nodeps: /Users/bullocm/git/aws-encryption-sdk-python/.tox/dist/aws-encryption-sdk-1.3.2.zip
flake8 installed: asn1crypto==0.23.0,attrs==17.2.0,aws-encryption-sdk==1.3.2,boto3==1.4.7,botocore==1.7.17,cffi==1.11.0,cryptography==2.0.3,docutils==0.14,flake8==3.4.1,flake8-docstrings==1.1.0,flake8-import-order==0.13,flake8-polyfill==1.0.1,idna==2.6,jmespath==0.9.3,mccabe==0.6.1,pycodestyle==2.3.1,pycparser==2.18,pydocstyle==2.0.0,pyflakes==1.5.0,python-dateutil==2.6.1,s3transfer==0.1.11,six==1.11.0,snowballstemmer==1.2.1,wrapt==1.10.11
flake8 runtests: PYTHONHASHSEED='1582386464'
flake8 runtests: commands[0] | flake8 src/aws_encryption_sdk/ setup.py
src/aws_encryption_sdk/internal/formatting/deserialize.py:63:1: C901 'deserialize_header' is too complex (15)
ERROR: InvocationError: '/Users/bullocm/git/aws-encryption-sdk-python/.tox/flake8/bin/flake8 src/aws_encryption_sdk/ setup.py'
pylint inst-nodeps: /Users/bullocm/git/aws-encryption-sdk-python/.tox/dist/aws-encryption-sdk-1.3.2.zip
[...]

PR #66, opened 2018-08-15
Pylint 2.0.0 and a little bit of housekeeping

mattsb42-aws:
*Issue #, if available:* #65 #17 #49 

*Description of changes:*
In the process of updating our code to comply with the new checks in Pylint 2.0.0, I made the additional changes:

- Finally got around to breaking `deserialize_header` logic out into component functions.
- Above enabled turning on flake8 and pylint runs in Travis.
- Turned on bandit run in Travis. 1.4.1 (or whatever the next version will be) still is not published, but we're pulling from their GitHub head for now.

### DIFF
--- a/src/aws_encryption_sdk/internal/formatting/deserialize.py
+++ b/src/aws_encryption_sdk/internal/formatting/deserialize.py
@@ -21,7 +21,7 @@ from cryptography.exceptions import InvalidTag
 
 from aws_encryption_sdk.exceptions import NotSupportedError, SerializationError, UnknownIdentityError
 from aws_encryption_sdk.identifiers import (
-    Algorithm, ContentType, ObjectType, SequenceIdentifier, SerializationVersion
+    AlgorithmSuite, ContentType, ObjectType, SequenceIdentifier, SerializationVersion
 )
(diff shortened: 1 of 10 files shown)


src/aws_encryption_sdk/internal/formatting/deserialize.py
```diff
@@ -61,109 +61,214 @@ def validate_header(header, header_auth, raw_header, data_key):
...
             'Content AAD length field is currently unused, its value must be always 0'
         )
-    header['content_aad_length'] = 0
 
-    (iv_length,) = unpack_values('>B', tee_stream)
-    if iv_length != alg.iv_len:
+    return 0
+
+
+def _verified_iv_length(iv_length, algorithm):
```
johnwalker: Probably makes sense to rename algorithm -> algorithm_suite. Same for any other parameters where this has changed.

mattsb42-aws: Fair point. Updated within the changed inner functions. I'm going to leave the rest as-is for now, though. At some point I would like to go through the codebase and do a more complete transition, but I think this is not the time to do that.

Issue #68, opened 2018-08-15
Add Python 3.7


mattsb42-aws:
Add Python 3.7 tests to CI and add 3.7 to supported versions.

PR #70, opened 2018-08-16
Add Python 3.7

mattsb42-aws:
*Issue #, if available:* #68 

*Description of changes:*
Add Python 3.7 to CI and advertised supported versions.


.travis.yml
```diff
@@ -34,6 +34,22 @@ matrix:
       env: TOXENV=py36-accept
     - python: 3.6
       env: TOXENV=py36-examples
+    - python: 3.7
+      env: TOXENV=py37-local
+      dist: xenial
```
lizroth: Is this required because 3.7 specifically? Can we doc or otherwise track the special case somehow?

mattsb42-aws: We should document it somewhere, yes. It's currently a necessary workaround for Travis' build configurations to get Python 3.7.

Issue #83, opened 2018-09-14
KMSMasterKeyProvider - requested master keys are generated before setting default region.


dtmistry:
When using KMSMasterKeyProvider, can a CMK key alias be used as the key_id instead of the arn?

The KMS API `GenerateDataKey` operation [accepts](https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html#API_GenerateDataKey_RequestSyntax) alias as the key_id. But it looks like the sdk expects the key_id to be an arn. If the key_id is not an arn, the init fails with the below error - 

The below fails with an exception - 

```python
kwargs = dict(
    key_ids=["alias/cmk-alias"],
    region_names=["us-east-1"],
    botocore_session=existing_session
)
aws_encryption_sdk.KMSMasterKeyProvider(**kwargs)      
```

```error
UnknownRegionError: No default region found and no region determinable from key id: alias/cmk-alias

/usr/local/lib/python2.7/site-packages/aws_encryption_sdk/key_providers/kms.py:50:UnknownRegionError
```

mattsb42-aws:
Thanks for reporting this!

It looks like this is caused by us finding the default region *after* generating the pre-defined master keys rather than *before*.

We'll get a fix out for this shortly, but in the meantime you can work around this by not pre-populating the key IDs.

```python
kwargs = dict(
    region_names=["us-east-1"],
    botocore_session=existing_session
)
key_provider = aws_encryption_sdk.KMSMasterKeyProvider(**kwargs)
key_provider.add_master_key("alias/cmk-alias")
```

mattsb42-aws:
This fix is now available in v1.3.7.

PR #84, opened 2018-09-14
fix KMSMasterKeyProvider to determine the default region before trying to create the requested master keys

mattsb42-aws:
*Issue #, if available:* #83

*Description of changes:*
Fix KMSMasterKeyProvider to determine the default region before trying to create the requested master keys.

We were previously trying to build the master keys *before* finding the default region, which would break any time a requested key ID was not a full Arn.


test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -25,6 +25,13 @@
 pytestmark = [pytest.mark.unit, pytest.mark.local]
 
 
+def test_init_with_regionless_key_ids_and_region_names():
+    key_ids = ("alias/key_1",)
+    region_names = ("test-reqion-1",)
```
karlw00t: reqion?

PR #87, opened 2018-09-27
Decrypt oracle

mattsb42-aws:
*Description of changes:*

This provides the initial iteration of the decrypt oracle.

Not addressed in this PR is how we plan to handle CD for this API. I'll tackle that separately.


decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/app.py
```diff
@@ -0,0 +1,68 @@
...
+from chalice import Chalice, Response
+
+from .key_providers.counting import CountingMasterKey
+from .key_providers.null import NullMasterKey
+
+CHALICE_DEBUG = os.environ.get("CHALICE_DEBUG", "no") == "yes"
+APP = Chalice(app_name="aws-encryption-sdk-decrypt-oracle", debug=CHALICE_DEBUG)
+APP.log.setLevel(logging.DEBUG)
+
+try:  # Python 3.5.0 and 3.5.1 have incompatible typing modules
```
karlw00t: this should only be building for 3.6.  Why do we need this?

decrypt_oracle/test/integration/integration_test_utils.py
```diff
@@ -0,0 +1,121 @@
...
+        raise ValueError(
+            'Environment variable "{}" must be set to a valid KMS CMK ARN for integration tests to run'.format(
+                AWS_KMS_KEY_ID
+            )
+        )
+
+    if arn.startswith("arn:") and ":alias/" not in arn:
+        return arn
+
+    raise ValueError("KMS CMK ARN provided for integration tests must be a key not an alias")
```
karlw00t: Can we raise a custom exception type here?

decrypt_oracle/README.rst
```diff
@@ -0,0 +1,45 @@
...
+that ciphertext can be successfully decrypted by the AWS Encryption SDK for Python.
+
+The API is deployed on Amazon API Gateway and backed by AWS Lambda.
+
+API v0
+======
+
+**Request**
+
+* **Method**: POST
```
karlw00t: what is the path on this URL request?

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/app.py
mattsb42-aws: That...is a fair point. I've gotten so much into the habit of having to support everything that I completely forgot that by the nature of the target environment this is strictly 3.6.

decrypt_oracle/test/integration/integration_test_utils.py
mattsb42-aws: Would that add value? The problem that this exception is identifying is that a user-provided value is incorrect. That seems to me to be the definition of `ValueError`.

decrypt_oracle/.chalice/policy-dev.json
```diff
@@ -0,0 +1,46 @@
...
+                "logs:CreateLogStream",
+                "logs:PutLogEvents"
+            ],
+            "Resource": [
+                "arn:aws:logs:*:*:log-group",
+                "arn:aws:logs:*:*:log-group:*"
+            ]
+        },
+        {
+            "Effect": "Deny",
```
SalusaSecondus: For this statement, I think you need to split it into two, because the NotAction and the NotResource are joined by an AND. So the statement as written says "DENY if and only if it is NOT an approved action AND it is NOT an approved resource." That means you won't get an explicit DENY if it is an unapproved action for an approved resource (or vice-versa).

decrypt_oracle/setup.py
```diff
@@ -0,0 +1,59 @@
...
+    keywords="aws-encryption-sdk aws kms encryption",
+    license="Apache License 2.0",
+    install_requires=get_requirements(),
+    classifiers=[
+        "Development Status :: 5 - Production/Stable",
+        "Intended Audience :: Developers",
+        "Natural Language :: English",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 2",
```
SalusaSecondus: Earlier you only list py36. Should all of these be listed here?

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/app.py
```diff
@@ -0,0 +1,72 @@
...
+    * 400 response code with whatever error code was encountered as the body
+    """
+    APP.log.debug("Request:")
+    APP.log.debug(json.dumps(APP.current_request.to_dict()))
+    APP.log.debug("Ciphertext:")
+    APP.log.debug(APP.current_request.raw_body)
+
+    try:
+        ciphertext = APP.current_request.raw_body
+        plaintext, _header = aws_encryption_sdk.decrypt(source=ciphertext, key_provider=_master_key_provider())
```
SalusaSecondus: Is there any way to return the `_header` information to callers as well?

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/key_providers/counting.py
```diff
@@ -0,0 +1,95 @@
...
+
+    provider_id = "test_counting"
+
+    def __init__(self) -> None:
+        """Set the key id to "test_counting_prov_info"."""
+        super(CountingMasterKeyConfig, self).__init__(key_id=b"test_counting_prov_info")
+
+
+class CountingMasterKey(MasterKey):
+    """Master key that generates deterministic data keys and decrypts a pre-defined
```
SalusaSecondus: Interesting piece of test code. Are we using it elsewhere?

It should come with a massive warning that it is *not* secure and must not be used for any non-test cases.

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/key_providers/null.py
```diff
@@ -0,0 +1,108 @@
...
+
+    provider_id = "null"
+
+    def __init__(self) -> None:
+        """Set the key id to "null"."""
+        super(NullMasterKeyConfig, self).__init__(key_id=b"null")
+
+
+class NullMasterKey(MasterKey):
+    """Master key that generates null data keys and decrypts any data key with provider id
```
SalusaSecondus: Interesting piece of test code. Are we using it elsewhere?

It should come with a massive warning that it is *not* secure and must not be used for any non-test cases.

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/key_providers/counting.py
mattsb42-aws: Not to my knowledge, no. Warning added in docstring that this should never be used outside of testing.

decrypt_oracle/src/aws_encryption_sdk_decrypt_oracle/key_providers/null.py
mattsb42-aws: Not to my knowledge, no. Warning added in docstring that this should never be used outside of testing.

decrypt_oracle/.chalice/policy-dev.json
mattsb42-aws: I'm not 100% on the policy combination, but splitting them into two (Action * with NotResource specifics and NotAction specifics with Resource *) made everything fail, including the CloudWatch logging. Reverted back to the (NotAction specifics with NotResource specifics) for now.

decrypt_oracle/test/integration/integration_test_utils.py
karlw00t: If you expect something to catch this exception, then yes.  Its cheap to provide a custom exception, and we aren't giving up anything that ValueError provides.

mattsb42-aws: The only place this error will manifest is in pytest error logs.

PR #91, opened 2018-10-16
First draft of keyrings for Python

mattsb42-aws:
*Description of changes:*

This is the first draft of when keyrings will look like in this client. I'm targeting a newly created feature branch because this is not yet ready to be merged into master.

This PR:
* introduces the required primitives that keyrings will handle
* introduces the base `Keyring` class and its interfaces
* updates `DefaultCryptographicMaterialsManager` to accept either a master key provider or a keyring
* adds a `MasterKeyKeyring` that provides a shim between master key providers and keyrings
  * This now contains the MK/P handling logic that was previously in the `DefaultCryptographicMaterialsManager`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ __pycache__
 
 # PyCharm
 .idea/
+venv/
 
 # Tox
 .tox
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,5 @@ boto3>=1.4.4
 cryptography>=1.8.1
 attrs>=17.4.0
 wrapt>=1.10.11
-zope.interface>=4.5.0
 six>=1.11.0
 enum34; python_version < '3.4'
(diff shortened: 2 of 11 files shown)


ttjsu-aws:
Looks to me that this implementation retains the Master Key Provider layer which we got rid of in the AWS Encryption SDK for C. In the AWS Encryption SDK for C we dissolved the MKP/MK abstractions and introduced the keyring abstraction.

mattsb42-aws:
@ttjsu-aws Yes, and we can do that (remove MK/P) in the new implementations, but the existing implementations need to be backwards compatible.

mattsb42-aws:
Closing this PR because the overall keyring design evolved fairly significantly since I wrote this.
I think it is best to simply start from scratch rather than iterating on this PR.

Issue #26, opened 2017-11-11
_EncryptionStream.read() does not properly handle b < 0


mattsb42-aws:
## Problem
`read()` for `b < 0` should readall. Currently, it will attempt to process a read of a negative value of bytes; it will only readall if `b is None`.

## Solution
Sanitize `b < 0` to `None`.

PR #98, opened 2018-11-08
Properly handle negative values to _EncryptionStream.read()

mattsb42-aws:
*Issue #, if available:* #26 

*Description of changes:*
The actual fix was tiny. Most of this PR is refactoring the test class from unittest to pytest.

### DIFF
--- a/src/aws_encryption_sdk/streaming_client.py
+++ b/src/aws_encryption_sdk/streaming_client.py
@@ -209,6 +209,10 @@ class _EncryptionStream(io.IOBase):
         :returns: Processed (encrypted or decrypted) bytes from source stream
         :rtype: bytes
         """
+        # Any negative value for b is interpreted as a full read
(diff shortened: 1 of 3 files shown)


src/aws_encryption_sdk/streaming_client.py
```diff
@@ -209,6 +209,10 @@ def read(self, b=None):
         :returns: Processed (encrypted or decrypted) bytes from source stream
         :rtype: bytes
         """
+        # Any negative value for b is interpreted as a full read
```
johnwalker: What is a full read?

mattsb42-aws: A full read means "read the entire message", and is triggered when either the number of bytes is not specified or when a negative value is specified.

https://docs.python.org/3/library/io.html#io.RawIOBase.read

Issue #24, opened 2017-11-01
streaming_client classes do not properly interpret short reads in source streams


mattsb42-aws:
## Problem
Per [PEP 3116](https://www.python.org/dev/peps/pep-3116/#raw-i-o):

> .read(n: int) -> bytes
> 
>     Read up to n bytes from the object and return them. Fewer than n bytes may be
>     returned if the operating system call returns fewer than n bytes. If 0 bytes are
>     returned, this indicates end of file. If the object is in non-blocking mode and no
>     bytes are available, the call returns None.

The current behavior of the clients in `streaming_clients` is to interpret reads returning fewer than n bytes as OEF.

## Solution
Update clients to properly handle short reads per PEP 3116.

PR #102, opened 2018-11-13
Fix handling of partial reads

mattsb42-aws:
*Issue #, if available:* #24 

*Description of changes:*
As noted in #24, the `streaming_client` classes have not historically correctly handled partial reads. Rather than accepting reads from the source stream that return fewer than the requested number of bytes, they assumed that this meant the source stream was complete. Per PEP 3116, this is incorrect: the stream should not be considered complete until no bytes are returned.

In order to simplify logic everywhere else, I ended up just adding a `ProxyObject` wrapper that keeps hitting the wrapped stream on read until it either gets the requested number of bytes or hits the end of the wrapped stream (as indicated by the stream returning nothing).

Tightly coupled with this was updating the fix made in #98 to correctly normalize to the bytes to read to `-1` rather than `None`. This required a few small tweaks throughout `streaming_client` to correctly handle this new value.

### DIFF
--- a/src/aws_encryption_sdk/streaming_client.py
+++ b/src/aws_encryption_sdk/streaming_client.py
@@ -223,7 +223,7 @@ class _EncryptionStream(io.IOBase):
         if self.closed:
             raise ValueError("I/O operation on closed file")
 
-        if b:
+        if b >= 0:
             self._read_bytes(b)
             output.write(self.output_buffer[:b])
             self.output_buffer = self.output_buffer[b:]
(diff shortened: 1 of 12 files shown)


src/aws_encryption_sdk/streaming_client.py
```diff
@@ -202,24 +202,28 @@ def readable(self):
         # Open streams are currently always readable.
         return not self.closed
 
-    def read(self, b=None):
+    def read(self, b=-1):
```
johnwalker: Curious, why change from None to -1?

src/aws_encryption_sdk/streaming_client.py
```diff
@@ -583,8 +600,8 @@ def _read_bytes(self, b):
         :param int b: Number of bytes to read
         :raises NotSupportedError: if content type is not supported
         """
-        _LOGGER.debug("%s bytes requested from stream with content type: %s", b, self.content_type)
-        if b <= len(self.output_buffer) or self.source_stream.closed:
+        _LOGGER.debug("%d bytes requested from stream with content type: %s", b, self.content_type)
+        if 0 <= b <= len(self.output_buffer) or self.source_stream.closed:
```
johnwalker: if b were None here, would 0 <= b <= len(x) return false?

src/aws_encryption_sdk/streaming_client.py
mattsb42-aws: PEP3116

https://www.python.org/dev/peps/pep-3116/

src/aws_encryption_sdk/streaming_client.py
mattsb42-aws: In Python 3 it would raise a `TypeError`. In Python 2, `None` is always less than any number.

However, in `read()` we normalize all `b` values of `None` to `-1`, and we test that `None` is accepted here[1] (added in #98 where I incorrectly normalized to `None`). The user interface is the same between #98 and here, but the internal implementation is more correct for PEP 3116.

[1] https://github.com/aws/aws-encryption-sdk-python/blob/master/test/unit/test_streaming_client_encryption_stream.py#L241

Issue #103, opened 2018-11-13
Reduce required API of source streams.


mattsb42-aws:
Currently, the source stream used by `streaming_client` streams must implement four APIs:

- `read()`
- `tell()`
- `close()`
- `closed`

As raised in #80, there are several cases where it would be useful for us to only require `read()`.

This leaves the following to be considered for elimination:

- [x] `tell()`
- [x] `close()`
- [x] `closed`

mattsb42-aws:
Current uses:

## tell()

- [x] `_EncryptionStream.stream_length()` : Failing here is ok. This just tells us that the stream is not seekable and we cannot determine the length of the source. We already have logic to handle that.
- [ ] `StreamDecryptor_prep_non_framed()` : Used to set the `body_start` value. We do not use this value anywhere, but unfortunately I made it public when I originally wrote this class. We need to consider whether we can just drop this attribute.
- [ ] `StreamDecryptor_read_bytes_from_non_framed_body()` : Used to determine how many more bytes to read. What we should do here instead is to read the unframed body metadata to determine how many bytes we need to read.

## close()

If we were implementing this new, I would say that we should not try to close the source stream at all.

- [ ] `StreamEncryptor._read_bytes_to_non_framed_body()` : Closes the source stream once we determine that we have as many bytes from it as we can expect.
- [ ] `StreamEncryptor._read_bytes_to_framed_body()` : Closes the source stream after we write the message footer.
- [ ] `StreamDecryptor._read_bytes_from_non_framed_body()` : Closes the source stream after we read the message footer.
- [ ] `StreamDecryptor._read_bytes_from_framed_body()` : Closes the source stream after we read the message footer.

## closed

- [ ] `_EncryptionStream.read()` : Used to keep reading until the source is closed. We can rework this instead just keep reading until we don't get anything back.
- [ ] `_EncryptionStream.next()` : Used to determine if we should `StopIteration`. I think that a better approach here would be to instead simply try reading and raise `StopIteration` if we either get nothing back or hit a `ValueError`.
- [ ] `StreamEncryptor._read_bytes()` : Used to short-circuit read if the source is closed. I think that we can safely ignore this. `InsistentReaderBytesIO` lets us work around any IO errors, and otherwise it saves us minimal processing time.
[...]

PR #104, opened 2018-11-14
Remove requirement for tell() on decrypt

mattsb42-aws:
*Issue #, if available:* #103 

*Description of changes:*
This is the first in a series of changes to implement #103.

This removes the need for the source stream to implement `tell()` on the decrypt path.

### DIFF
--- a/src/aws_encryption_sdk/streaming_client.py
+++ b/src/aws_encryption_sdk/streaming_client.py
@@ -767,7 +767,7 @@ class StreamDecryptor(_EncryptionStream):  # pylint: disable=too-many-instance-a
 
     def _prep_non_framed(self):
         """Prepare the opening data for a non-framed message."""
-        self._unframed_body_iv, self.body_length = aws_encryption_sdk.internal.formatting.deserialize.deserialize_non_framed_values(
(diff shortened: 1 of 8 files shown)


mattsb42-aws:
Rebased on master to catch `1.3.8` changes.

src/aws_encryption_sdk/streaming_client.py
```diff
@@ -751,9 +753,21 @@ def _read_header(self):
         )
         return header, header_auth
 
+    @property
+    def body_start(self):
+        """Log deprecation warning when body_start is accessed."""
+        _LOGGER.warning("StreamDecryptor.body_start is deprecated and will be removed in 1.4.0")
```
johnwalker: Would adding this deprecation message be released through a minor rev or a patch rev?

mattsb42-aws: Per our versioning statement, the warnings should be added in a `Z` rev and the removal should happen in a `Y` rev.

https://github.com/aws/aws-encryption-sdk-python/blob/master/VERSIONING.rst

PR #110, opened 2018-12-10
Remove requirement for tell() on source_stream

mattsb42-aws:
*Issue #, if available:* #103 

*Description of changes:*
Remove the requirement for `source_stream.tell()`.

### DIFF
--- a/src/aws_encryption_sdk/streaming_client.py
+++ b/src/aws_encryption_sdk/streaming_client.py
@@ -399,6 +399,8 @@ class StreamEncryptor(_EncryptionStream):  # pylint: disable=too-many-instance-a
         ):
             raise SerializationError("Source too large for non-framed message")
 
+        self.__unframed_plaintext_cache = io.BytesIO()
+
     def ciphertext_length(self):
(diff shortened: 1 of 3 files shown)


src/aws_encryption_sdk/streaming_client.py
```diff
@@ -486,14 +488,25 @@ def _write_header(self):
 
     def _prep_non_framed(self):
         """Prepare the opening data for a non-framed message."""
+        try:
+            plaintext_length = self.stream_length
```
johnwalker: Is a streaming client created per stream, or is it thread safe? Curious how instance variables work in these objects.

mattsb42-aws: Streaming client instances are not threadsafe. Instances should not be shared across threads.

Class attributes defined and modified in the class definition (we don't have any of those) are global across all instances of that class. Any attributes defined or modified on the instance are local to that instance.

johnwalker: Gotcha sounds good.

Issue #99, opened 2018-11-13
Convert all tests to pytest


mattsb42-aws:
A lot of the original test suite was written in `unittest`. Let's get all of that moved over to `pytest`.

Initially, this should be converting `unittest.TestCase` classes to `pytest` classes similar to what was done in #98. We can pull those methods out to stand-alone functions later if we decide we want to, but leaving them in classes will make the diffs most simpler.

Test modules to convert:

- [ ] test/integration/test_i_aws_encrytion_sdk_client.py:72:class TestKMSThickClientIntegration(unittest.TestCase):
- [ ] test/unit/test_aws_encryption_sdk.py:26:class TestAwsEncryptionSdk(unittest.TestCase):
- [ ] test/unit/test_defaults.py:23:class TestDefaults(unittest.TestCase):
- [ ] test/unit/test_deserialize.py:32:class TestDeserialize(unittest.TestCase):
- [ ] test/unit/test_encryption_context.py:29:class TestEncryptionContext(unittest.TestCase):
- [ ] test/unit/test_providers_base_master_key.py:66:class TestMasterKey(unittest.TestCase):
- [ ] test/unit/test_providers_base_master_key_provider.py:66:class TestBaseMasterKeyProvider(unittest.TestCase):
- [ ] test/unit/test_providers_kms_master_key.py:33:class TestKMSMasterKey(unittest.TestCase):
- [ ] test/unit/test_providers_kms_master_key_provider.py:35:class TestKMSMasterKeyProvider(unittest.TestCase):
- [ ] test/unit/test_providers_raw_master_key.py:30:class TestRawMasterKey(unittest.TestCase):
- [ ] test/unit/test_providers_raw_master_key_provider.py:45:class TestRawMasterKeyProvider(unittest.TestCase):
- [ ] test/unit/test_serialize.py:54:class TestSerialize(unittest.TestCase):
- [ ] test/unit/test_streaming_client_stream_decryptor.py:32:class TestStreamDecryptor(unittest.TestCase):
- [ ] test/unit/test_streaming_client_stream_encryptor.py:39:class TestStreamEncryptor(unittest.TestCase):
- [ ] test/unit/test_util_str_ops.py:25:class TestStrOps(unittest.TestCase):
- [ ] test/unit/test_utils.py:50:class TestUtils(unittest.TestCase):

AndresSan6:
Helllo! I can help with this issue, I will start working on it.

[...]

PR #127, opened 2018-12-14
Migrate unit/test_streaming_client_stream_encryptor.py from unittest to pytest

AndresSan6:
*Issue #, if available:* #99

*Description of changes:*
Migrated "test/unit/test_streaming_client_stream_encryptor.py" from unittest to pytest

### DIFF
--- a/test/unit/test_streaming_client_stream_encryptor.py
+++ b/test/unit/test_streaming_client_stream_encryptor.py
@@ -12,7 +12,6 @@
 # language governing permissions and limitations under the License.
 """Unit test suite for aws_encryption_sdk.streaming_client.StreamEncryptor"""
 import io
-import unittest
 
 import pytest
 import six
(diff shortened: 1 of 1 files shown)


test/unit/test_streaming_client_stream_encryptor.py
```diff
@@ -36,8 +35,9 @@
 pytestmark = [pytest.mark.unit, pytest.mark.local]
 
 
-class TestStreamEncryptor(unittest.TestCase):
-    def setUp(self):
+class TestStreamEncryptor(object):
+    @pytest.fixture(autouse=True)
+    def apply_fixtures(self):
```
mattsb42-aws: We also need to run the old `tearDown` actions after the tests are done. This is what is likely causing most of the test failures.

This should work by just running them in this method after a yield.

```python
@pytest.fixture(autouse=True)
def apply_fixtures(self):
    do_stuff_from_old_setup()
    yield
    do_stuff_from_old_teardown()
```

https://docs.pytest.org/en/latest/fixture.html#autouse-fixtures-xunit-setup-on-steroids

test/unit/test_streaming_client_stream_encryptor.py
```diff
@@ -487,8 +488,9 @@ def test_read_bytes_unsupported_type(self, mock_read_non_framed, mock_read_frame
         test_encryptor._encryption_materials = self.mock_encryption_materials
         test_encryptor._header = MagicMock()
         test_encryptor.content_type = None
-        with six.assertRaisesRegex(self, NotSupportedError, "Unsupported content type"):
+        with pytest.raises(NotSupportedError) as excinfo:
             test_encryptor._read_bytes(5)
+        execinfo.match("Unsupported content type")
```
mattsb42-aws: Typo:

```suggestion
        excinfo.match("Unsupported content type")
```

PR #144, opened 2019-03-18
Tweak example text

juneb:

### DIFF
--- a/examples/src/data_key_caching_basic.py
+++ b/examples/src/data_key_caching_basic.py
@@ -47,7 +47,7 @@ def encrypt_with_caching(kms_cmk_arn, max_age_in_cache, cache_capacity):
 
     # When the call to encrypt data specifies a caching CMM,
     # the encryption operation uses the data key cache specified
-	# in the caching CMM
+    # in the caching CMM
     encrypted_message, _header = aws_encryption_sdk.encrypt(
         source=my_data, materials_manager=caching_cmm, encryption_context=encryption_context
     )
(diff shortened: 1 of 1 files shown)


examples/src/data_key_caching_basic.py
```diff
@@ -45,8 +45,9 @@ def encrypt_with_caching(kms_cmk_arn, max_age_in_cache, cache_capacity):
         max_messages_encrypted=MAX_ENTRY_MESSAGES,
     )
 
-    # When the call to encryptData specifies a caching CMM,
-    # the encryption operation uses the data key cache
+    # When the call to encrypt data specifies a caching CMM,
+    # the encryption operation uses the data key cache specified
+	# in the caching CMM
```
mattsb42-aws: You got tabs in my spaces! :)

https://travis-ci.org/aws/aws-encryption-sdk-python/jobs/507932574

david-koenig: @juneb : The issue is that the comments on lines 48 and 49 have 4 spaces before the # symbol, but on line 50 there is a single tab character instead of 4 spaces before the # symbol. This is causing Travis CI to reject this file.

Different text editors render the tab in different ways, so in your text editor you may not see any misalignment between this line and the other ones. For fixing this individual problem, you can probably just delete the indent and then put four spaces in its place. More generally, your text editor probably has a setting somewhere to "use spaces instead of tabs" that will stop this problem from recurring.

juneb: Thanks, @david-koenig. Fixed in file and in my editor.

PR #145, opened 2019-03-25
Fix small doc typos.

lizroth:
*Issue #, if available:* n/a

*Description of changes:* Spotted a couple of typos while rereading the code and thought I would tidy up.


mattsb42-aws:
Travis appears to be in a weird hung state after the merge, and I'm confident this is not introducing anything that would break things, so I'm going to override that requirement.

mattsb42-aws:
:/ nvm then...apparently you can't override requirements, which is in general a good thing. I'll poke at Travis and see if we can't get it unstuck.

Issue #156, opened 2019-05-24
Revise Examples 


ttjsu-aws:
The AWS Encryption SDK Examples in Python is not intuitive/simple to understand. The API usage is not immediately apparent and is buried in a lot of details. 

It would be useful to have examples that are a combination of the commonly used APIs, algorithm suites and the type of data (string/stream) to be encrypted/decrypted.

**A. Commonly used APIs :**

1. Encrypt/Decrypt using one KMS CMK
2. Encrypt/Decrypt using multiple KMS CMKs in Multiple Regions
3. Decrypt using KMS CMK with the client provider 
4. Decrypt using KMS Discovery with region limits/filters
5. Decrypt using KMS generic discovery
6. Encrypt/Decrypt using Multi-KR/Multi-MKP
7. Caching 1:1 local cache: caching cmm
8. Caching 1:n local cache: caching cmm

Once we have the Keyring Implementation: https://github.com/aws/aws-encryption-sdk-python/issues/146 

9.  Encrypt/ Decrypt using Raw AES Keyring 
10.  Encrypt using Raw RSA Keyring (public-key) 
11.  Decrypt using Raw RSA Keyring (public and private) 
12.  Encrypt/Decrypt using KMS Keyring
13.  Encrypt/Decrypt using  Multi-Keyring

**B. Types of data for encryption/decryption:**

1. One shot/ string 
2. Streaming data

**C. Commonly used Algorithm Suite:**

1. Default Algorithm Suite AES_256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384 with P-384 and SHA-384
2. Default w/o signing AES256_GCM_IV12_TAG16_HKDF_SHA256

Examples: 

- [x]     A1.B1.C1 #236
- [x]     A1.B2.C1 #236
- [x]     A1.B1.C2 #236
- [x]     A2.B1.C1 #177
- [x]     A3.B1.C1 #236
- [x]     A4.B1.C1 #236
- [x]     A5.B1.C1 #236
- [x]     A6.B1.C1 #236
- [x]     A7.B1.C1 #239
- [x]     A8.B1.C1 #239
- [x]     A9.B1.C1 #221
- [x]     A10.B1.C1 #221
- [x]     A11.B1.C1 #221
- [x]     A12.B1.C1 #221
- [x]     A13.B1.C1 #221

 
Also, be sure to include/update the README to help navigate through the examples.

mattsb42-aws:
[...]

PR #157, opened 2019-05-29
reorganize examples cross-imports to simplify tests

mattsb42-aws:
*Description of changes:*
In preparation for #156, I realized that I hadn't given these examples tests the same treatment that I gave the DDB encryption client and Pipeformer.

These changes make writing tests for examples much simpler and less finicky.


mattsb42-aws:
It looks like the AppVeyor CPython34-x64 runner is still failing. :( Best we've been able to tell, this is a bug on their end. We need to figure that out eventually, but not necessarily now.

PR #158, opened 2019-06-04
Adding example one_kms_cmk

MeghaShetty:
*Issue #, if available:*

*Description of changes:*


examples/test/test_i_one_kms_cmk.py
```diff
@@ -0,0 +1,33 @@
...
+import os
+
+import botocore.session
+import pytest
+
+from ..src.one_kms_cmk import encrypt_decrypt
+from .examples_test_utils import get_cmk_arn
+
+
+pytestmark = [pytest.mark.examples]
```
karlw00t: Does this need to be #noqa markered? Or does it need to be deleted?

examples/test/test_i_one_kms_cmk.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_one_kms_cmk():
+    plaintext = os.urandom(1024)
```
karlw00t: It's generally bad form to have random data used in a unit test. The reason being that if this fails, it will fail once, due to some randomness. We should make our unit test deterministic

examples/test/test_i_one_kms_cmk.py
mattsb42-aws: Nope. Pylint and flake8 know to ignore these. I forget if that's because they have an understanding of pytest built in or because pytest uses it implicitly.

examples/test/test_i_one_kms_cmk.py
mattsb42-aws: I'm ok with this either way, tbh. In the case of the examples, the input plaintext is purely arbitrary, so it can literally be anything.

That said, if we want to change this to define a static plaintext, we should probably define it once in `examples_test_utils` and use it in all of the examples.

examples/src/one_kms_cmk.py
```diff
@@ -0,0 +1,31 @@
+import aws_encryption_sdk
```
mattsb42-aws: We need the copyright notice and a docstring for the file.

examples/src/one_kms_cmk.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def encrypt_decrypt(key_arn, source_plaintext, botocore_session=None):
```
mattsb42-aws: We need a docstring for this function.

examples/test/test_i_one_kms_cmk.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2017-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mattsb42-aws: This should be "Copyright 2019", not "2017-2019".

The year/range defines two things: when the file was created and when it was most recently edited.
If those years are the same, we do not need a range.

examples/test/test_i_one_kms_cmk.py
```diff
@@ -0,0 +1,33 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Unit test suite for the Strings examples in the AWS-hosted documentation."""
```
mattsb42-aws: This is inaccurate. A reasonable description of this would be to identify the examples file that this test tests.

MeghaShetty: I will make all the above changes. Thank you!

examples/test/examples_test_utils.py
```diff
@@ -17,4 +17,8 @@
...
...bulum non ligula tincidunt, finibus sapien in, sollicitudin ex. Pellentesque congue laoreet mi in condimentum. Cras convallis nisi ac nunc tincidunt venenatis. Suspendisse urna elit, cursus eu lacus a, aliquet porttitor mi. Nulla vel congue nibh, sed condimentum dui. Ut ante ligula, blandit eu finibus nec, scelerisque quis eros. Maecenas gravida odio eget nibh dictum, dictum varius lacus interdum. Integer quis nulla vulputate, rhoncus diam vitae, mollis mauris. Sed ut porttitor dolor. Fusce ut justo a ex bibendum imperdiet nec sit amet magna. Sed ullamcorper luctus augue, tempor viverra elit interdum sed. Cras sit amet arcu eu turpis molestie sollicitudin. Curabitur fermentum varius nibh, ut aliquet nisi. Aliquam id tempus tellus. Nulla porttitor nulla at nibh interdum, quis sollicitudin erat egestas. Ut blandit mauris quis efficitur efficitur. Morbi neque sapien, posuere ut aliquam eget, aliquam at velit. Morbi sit amet rhoncus felis, et hendrerit sem. Nulla porta dictum ligula eget iaculis. Cras lacinia ligula quis risus ultrices, sed consectetur metus imperdiet. Nullam id enim vestibulum nibh ultricies auctor. Morbi neque lacus, faucibus vitae commodo quis, malesuada sed velit.'
```
mattsb42-aws: Please wrap this to >=120 character lines.

You can do this easily with strings using [string literal concatenation](https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation).

ex:
```python
static_plaintext = (
    b"aosiejfoiawejfoiawj"
    b"aisjefoiawejsjowieja"
    ...
)
```

examples/test/examples_test_utils.py
```diff
@@ -17,4 +17,8 @@
...
+static_plaintext = str.encode(static_plaintext)
```
mattsb42-aws: We're defining the value manually; just make it bytes to begin with.

```python
# This is a string
static_plaintext = "some data"

# This is some bytes
static_plaintext = b"some data"
```

examples/test/test_i_basic_encryption.py
```diff
@@ -11,19 +11,18 @@
...
 # language governing permissions and limitations under the License.
 """Unit test suite for the Strings examples in the AWS-hosted documentation."""
-import os
-
 import botocore.session
 import pytest
 
 from ..src.basic_encryption import cycle_string
 from .examples_test_utils import get_cmk_arn
+from .examples_test_utils import static_plaintext
```
mattsb42-aws: We prefer to have all imports from the same module in the same import statement.

examples/test/test_i_one_kms_cmk_streaming_data.py
```diff
@@ -0,0 +1,41 @@
...
+from .examples_test_utils import get_cmk_arn
+from .examples_test_utils import static_plaintext
+
+
+pytestmark = [pytest.mark.examples]
+
+
+def test_one_kms_cmk_streaming_data():
+    cmk_arn = get_cmk_arn()
+    handle, filename = tempfile.mkstemp()
```
mattsb42-aws: We don't use `handle`, so we should be explicit about that in its naming.

https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

```suggestion
    _handle, filename = tempfile.mkstemp()
```

mattsb42-aws:
The AppVeyor failures are expected, unfortunately. We need to figure out why (#159), but they're failing across the board, not just for this PR.

examples/test/test_i_one_kms_cmk_streaming_data.py
```diff
@@ -0,0 +1,40 @@
...
+    with open(filename, "wb") as f:
+        f.write(static_plaintext)
+    try:
+        new_files = encrypt_decrypt_stream(
+            key_arn=cmk_arn, source_plaintext_filename=filename, botocore_session=botocore.session.Session()
+        )
+        for f in new_files:
+            os.remove(f)
+    finally:
+        os.close(_handle)
```
mattsb42-aws: Whoops, that was my mistake. I missed that we do actually use this. Sorry about that.
```suggestion
        os.close(handle)
```

examples/test/test_i_one_kms_cmk_streaming_data.py
```diff
@@ -0,0 +1,40 @@
...
+from ..src.one_kms_cmk_streaming_data import encrypt_decrypt_stream
+from .examples_test_utils import get_cmk_arn, static_plaintext
+
+
+pytestmark = [pytest.mark.examples]
+
+
+def test_one_kms_cmk_streaming_data():
+    cmk_arn = get_cmk_arn()
+    _handle, filename = tempfile.mkstemp()
```
mattsb42-aws:
```suggestion
    handle, filename = tempfile.mkstemp()
```

MeghaShetty: Changed it back

PR #163, opened 2019-06-21
Keyring materials

mattsb42-aws:
*Description of changes:*

Update Encryption/Decryption materials and create Keyring traces in preparation for Keyring work.


src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -51,28 +52,87 @@ class EncryptionMaterialsRequest(object):
...
+@attr.s
+class CryptographicMaterials(object):
+    """Cryptographic materials core.
+
+    .. versionadded:: 1.5.0
+
+    :param Algorithm algorithm: Algorithm to use for encrypting message
+    :param dict encryption_context: Encryption context tied to `encrypted_data_keys`
+    :param DataKey data_encryption_key: Plaintext data key to use for encrypting message
+    :param encrypted_data_keys: List of encrypted data keys
```
seebees: I think that encrypted_data_keys should be in `EncryptionMaterials`.  Since Decryption does not need the EDK...

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -92,16 +152,64 @@ class DecryptionMaterialsRequest(object):
...
+    :type keyring_trace: list of :class:`KeyringTrace`
     """
 
-    data_key = attr.ib(validator=attr.validators.instance_of(DataKey))
     verification_key = attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(bytes)))
+
+    def __init__(self, data_key=_DEFAULT_SENTINEL, verification_key=None, **kwargs):
+        if any(
+            (
+                data_key is _DEFAULT_SENTINEL and "data_encryption_key" not in kwargs,
```
seebees: Is this where we exclude the EDK and the data key?

mattsb42-aws: No, this is to maintain backwards compatibility with the old `DecryptionMaterials` API.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: Thinking about this, I'm inclined to agree with you on this. I wanted to keep everything in one piece, just like the `on_encrypt` path, but I agree that once the decryption materials leave the Keyring layer, the EDKs are not needed.

mattsb42-aws:
@seebees @lizroth Something that occurred to me working through this:

The way I am maintaining the keyring trace currently is via a list containing `keyring identifier / keyring trace flags` pairs. However, if a keyring generates the data key *and* adds an encrypted data key, then we end up with two entries in that list.

I'm inclined to say now that it would be better to instead maintain this as a map of `keyring identifier` to `keyring trace flags`. That way it would be more straightforward to locate all of the flags added by a given keyring.

Thoughts?

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,250 @@ class EncryptionMaterialsRequest(object):
...
-    :type encrypted_data_keys: list of `aws_encryption_sdk.structures.EncryptedDataKey`
+    .. versionadded:: 1.5.0
+
+        The **keyring_trace** parameter.
+
+    .. versionadded:: 1.5.0
+
+        Most parameters are now optional.
+
+    :param Algorithm algorithm: Algorithm to use for encrypting message
```
seebees: tiny nit, are we not moving to "Algorithm Suite"?

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,250 @@ class EncryptionMaterialsRequest(object):
...
+        default=attr.Factory(list), validator=optional(deep_iterable(member_validator=instance_of(EncryptedDataKey)))
+    )
+    signing_key = attr.ib(default=None, repr=False, validator=optional(instance_of(bytes)))
+
+    def __init__(
+        self,
+        algorithm=None,
+        data_encryption_key=None,
+        encrypted_data_keys=None,
+        encryption_context=None,
```
seebees: I chose not to include the Encryption Context in the material.  I did this because the top level caller _must_ have the encryption context.  I'm not saying we should remove it, just that it is slightly different.

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,250 @@ class EncryptionMaterialsRequest(object):
...
+        :raises InvalidKeyringTraceError: if keyring trace does not match data key provider
+        :raises InvalidDataKeyError: if data key length does not match algorithm suite
+        """
+        self._add_data_encryption_key(
+            data_encryption_key=data_encryption_key,
+            keyring_trace=keyring_trace,
+            required_flags={KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY},
+        )
+
+    def add_encrypted_data_key(self, encrypted_data_key, keyring_trace):
```
seebees: I would also add a check to make sure the `data_encryption_key` has been set.
https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts#L358

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: This was to maintain backwards compatibility with the old `EncryptionMaterials` API, but you're right. I think adding something like the `DecryptionMaterials.data_key` redirect is the right thing to do here.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: Yes, I like that.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: Actually, looking back through the API, I thought we had moved to `algorithm_suite` elsewhere but we have not; it looks like the only other places we have moved to that name is in internal APIs.

For consistency within this client, I'm going to leave this as-is.

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,254 @@ class EncryptionMaterialsRequest(object):
...
+            raise InvalidKeyringTraceError("Keyring trace does not match data key provider.")
+
+        if len(data_encryption_key.data_key) != self.algorithm.kdf_input_len:
+            raise InvalidDataKeyError(
+                "Invalid data key length {actual} must be {expected}.".format(
+                    actual=len(data_encryption_key.data_key), expected=self.algorithm.kdf_input_len
+                )
+            )
+
+    def _add_data_encryption_key(self, data_encryption_key, keyring_trace, required_flags):
```
seebees: I keep reading this name and confusing it.

`data_encryption_key` and `encrypted_data_key` seem a little close?

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,254 @@ class EncryptionMaterialsRequest(object):
...
+        :raises AttributeError: if signing key is already set
+        :raises SignatureKeyError: if algorithm suite does not support signing keys
+        """
+        if self.signing_key is not None:
+            raise AttributeError("Signing key is already set.")
+
+        if self.algorithm.signing_algorithm_info is None:
+            raise SignatureKeyError("Algorithm suite does not support signing keys.")
+
+        self._setattr("signing_key", signing_key)
```
seebees: We should check that the signatureCurve for the key is the same as the suite....

https://github.com/awslabs/aws-encryption-sdk-javascript/issues/130

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -87,21 +311,101 @@ class DecryptionMaterialsRequest(object):
...
+
+        self._setattr("verification_key", verification_key)
+        attr.validate(self)
+
+    @property
+    def data_key(self):
+        """Backwards-compatible shim for access to data key."""
+        return self.data_encryption_key
+
+    def add_data_encryption_key(self, data_encryption_key, keyring_trace):
```
seebees: Don't we need a `_validate_data_encryption_key` call here?

mattsb42-aws: That's strictly a shim for access via the old API. We put guardrails around changing the data encryption key and we check it when you set it through those guardrails. I don't think there's really much benefit checking it *again* in this one specific access path.

If we want to validate the data encryption key on access, we should do that via a `property` pattern for `CryptographicMaterials.data_encryption_key`.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: You're not wrong, but that's what the API has been since we introduced CMMs. Either we need to change the API with backwards compatible shims (see `DecryptionMaterials`) or we can just accept the similar names.

seebees: sadness

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: I agree, but digging into this, I think the right way to address this will be my adding this verification into the `Signer` and `Verifier` classes that we use to abstract all of the signing and verification away. For the time being, I'm just going to create an instance of each to "verify" the key and address the actual verification in https://github.com/aws/aws-encryption-sdk-python/issues/167.

src/aws_encryption_sdk/identifiers.py
```diff
@@ -328,3 +328,13 @@ class ContentAADString(Enum):
     FRAME_STRING_ID = b"AWSKMSEncryptionClient Frame"
     FINAL_FRAME_STRING_ID = b"AWSKMSEncryptionClient Final Frame"
     NON_FRAMED_STRING_ID = b"AWSKMSEncryptionClient Single Block"
+
+
+class KeyringTraceFlag(Enum):
```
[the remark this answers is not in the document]
seebees: JS does the bit flag thing because I copied C.  I don't much like the bit flag because people will not understand it.
While I can get behind your "specification" point, I'm not sure that this is strictly a specification thing.  Knowing what wrapping key did what should be, but how this is recorded?

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +48,254 @@ class EncryptionMaterialsRequest(object):
...
+        data_key = _data_key_to_raw_data_key(data_key=data_encryption_key)
+
+        super(CryptographicMaterials, self).__setattr__("data_encryption_key", data_key)
+        self._keyring_trace.append(keyring_trace)
+
+    @property
+    def keyring_trace(self):
+        """Return a read-only version of the keyring trace.
+
+        :rtype: tuple
```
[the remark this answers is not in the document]
mattsb42-aws: What I'm trying to do here (and with `Encryptionmaterials.encrypted_data_keys`) is to remove the easy path for people to modify the encrypted data keys or keyring trace directly without going through the `add_` methods.

The reason for this is to make it harder for people to add one without the other.

src/aws_encryption_sdk/identifiers.py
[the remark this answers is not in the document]
mattsb42-aws: We haven't touched on this yet in the spec, no.

My general take on this is that we have not really been defining this as a thing that should be passed around between implementations, but instead have been defining it as a thing that is used as a unique reference within an implementation.

Talking with the folks who worked on the C client, the reason for using the bitshift values was for simple storage and comparison in a world where you lack things like sets.

I guess it wouldn't hurt anything to set the values to those bit-shift values just in case we want to use them later, and make the decision on if we do or not when we get to that point with the spec.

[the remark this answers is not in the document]
mattsb42-aws: >  While I can get behind your "specification" point, I'm not sure that this is strictly a specification thing. Knowing what wrapping key did what should be, but how this is recorded?

I think the question is more: "is this a value that we want to expose in any artifact that might cross implementations?"

We have precedent for values like this in, for example, the cache entry identifier and encryption context sorting order.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: No, this was addressed in a later commit on this PR....but it looks like I forgot to update the docs... :/  I'll do that.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: IMO we should include the encryption context so that it is more directly bound.

Specifically, the encryption context that needs to be used as part of these encryption materials, as possibly distinct from the encryption context that was written in the encrypted message.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: Unless this is causing customer confusion, I don't think so. This would be a non-trivial change for (IMO) not much benefit.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: merged in https://github.com/aws/aws-encryption-sdk-python/pull/163/commits/b311cda30958283860f198f2f61a24d928af6d9e

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: fixed in https://github.com/aws/aws-encryption-sdk-python/pull/163/commits/10ded57e01e1e4d130860ae94c10f5b7db397471

src/aws_encryption_sdk/identifiers.py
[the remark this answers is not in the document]
seebees: > We have precedent for values like this...

Then I vote bit flags all the way down.

seebees:
>I'm inclined to say now that it would be better to instead maintain this as a map of keyring identifier to keyring trace flags. That way it would be more straightforward to locate all of the flags added by a given keyring.

I agree.

https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts#L371

src/aws_encryption_sdk/materials_managers/__init__.py
seebees: >encryption context that needs to be used as part of these encryption materials
That is the part that convinces me :)

I can get down with this.  The part I don't like is that the material is created by the CMM, when is the encryption context set?  The _best_ time would be when the material is created.  Otherwise I have to work to distinguish between "no encryption context" and "encryption context has not yet been set so these material are not yet valid"

src/aws_encryption_sdk/materials_managers/__init__.py
```diff
@@ -40,38 +49,285 @@ class EncryptionMaterialsRequest(object):
...
+
+        if not self.encrypted_data_keys:
+            return False
+
+        if self.algorithm.signing_algorithm_info is not None and self.signing_key is None:
+            return False
+
+        return True
+
+    def add_data_encryption_key(self, data_encryption_key, keyring_trace):
```
seebees: Re: https://github.com/aws/aws-encryption-sdk-python/pull/163#discussion_r299256972

I think a comment about _why_ the keyring trace is here?  Maybe it will resolve the other convo?

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: >  The part I don't like is that the material is created by the CMM

This is explicitly part of the CMM contract: any CMM MAY modify the encryption context in either direction.

> when is the encryption context set?

On initial creation of materials. The keyring MUST not modify the encryption context.

mattsb42-aws:
> > I'm inclined to say now that it would be better to instead maintain this as a map of keyring identifier to keyring trace flags. That way it would be more straightforward to locate all of the flags added by a given keyring.
>
> I agree.
>
> https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts#L371

I've actually been oscillating the other way on this. Mainly because: if a keyring does the same action twice, how should that be recorded in the trace?

We can leave this for resolution after this PR; the write API on the materials will not change, just the read API, which will not affect the other ongoing keyring development.

src/aws_encryption_sdk/identifiers.py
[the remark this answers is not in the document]
mattsb42-aws: done

src/aws_encryption_sdk/materials_managers/__init__.py
seebees: Good, I'm cool with "initial creation of materials".  But once a CMM creates materials, downstream CMM's can ~not change the Encryption Context...  Or they have to re-create materials and clone things over.

i.e. this make is tedious to send an encryption context to the KMS keyring, but *not* return that encryption context to be serialized into the message format.

mattsb42-aws:
Looks like something in `pydocstyle 4.0.0` broke `flake8-docstrings`. For the moment I'm going to put in a temporary pin to `<4.0.0`. We can deal with that separately.

mattsb42-aws:
Per offline conversation, changing `EncryptionMaterials.encrypted_data_keys` away from set/frozenset to list/tuple.

src/aws_encryption_sdk/materials_managers/__init__.py
mattsb42-aws: >  Or they have to re-create materials and clone things over.

This. CMMs MAY modify the encryption context. They also MAY return whatever encryption/decryption materials that they deem appropriate. That MAY include a CMM layer that creates new materials based in part, whole, or not at all, on the materials returned from an underlying CMM.

mattsb42-aws:
AppVeyor is failing install on 2.7 and 3.4, but that looks unrelated to these changes to me so since this is not going into master yet anyway I'm inclined to merge this now to unblock the rest of the keyring work.

PR #161, opened 2019-06-18
Keyring base API

MeghaShetty:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #165, opened 2019-06-27
Raw keyrings

MeghaShetty:
*Issue #, if available:*

*Description of changes:*
Dependent on the merging of the PRs for base API #161  and Keyring materials #163 .


src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    if decryption_materials.data_key:
+        return decryption_materials
+
+    # Wrapped EncryptedDataKey to deserialized EncryptedData
+    encrypted_wrapped_key = aws_encryption_sdk.internal.formatting.deserialize.deserialize_wrapped_key(
+        wrapping_algorithm=wrapping_algorithm, wrapping_key_id=key_name, wrapped_encrypted_key=encrypted_data_key
+    )
+
+    # EncryptedData to raw key string
+    plaintext_data_key = wrapping_key.decrypt(
```
mattsb42-aws: Wrap this decrypt call with a try/except catching all exceptions. In the `except` block, log the exception type and message, return the decryption materials unchanged.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+        encrypted_wrapped_data_key=encrypted_wrapped_key, encryption_context=decryption_materials.encryption_context
+    )
+
+    # Update decryption materials
+    data_encryption_key = DataKey(
```
mattsb42-aws: This should be a `RawDataKey`. There's no need for the encrypted version past here.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+        key_provider=encrypted_data_key.key_provider,
```
mattsb42-aws: This should be a new `MasterKeyInfo` created using this keyring. This is especially important for the `RawAESKeyring` since key name and provider info are different for that keyring.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
+    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(six.binary_type))
+    _wrapping_key = attr.ib(hash=True, repr=False, validator=attr.validators.instance_of(WrappingKey))
+    _wrapping_algorithm = attr.ib(repr=False, validator=attr.validators.instance_of(WrappingAlgorithm))
+
+    def __attrs_post_init__(self):
+
+        _key_provider = MasterKeyInfo(provider_id=self.key_namespace, key_info=self.key_name)
+
+        _key_info_prefix = struct.pack(
```
mattsb42-aws: Rather than duplicating this logic here, go ahead and pull the logic from `aws_encryption_sdk.internal.formatting.serialize.serialize_raw_master_key_prefix` into a lower-level helper function (that takes a key name and algorithm suite as input) that `serialize_raw_master_key_prefix` will then call, and use that lower-level helper function directly here.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+        :param encryption_materials: Contains signing key, encryption context and algorithm suite
+                                    required to encrypt data key
+        :type encryption_materials: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :returns: Contains signing key, unencrypted data key, encrypted data keys,
+                                    encryption context and algorithm suite required to encrypt data key
+        :rtype: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        """
+
+        encryption_materials = on_encrypt_helper(
+            encryption_materials, self._key_provider, self._wrapping_key, self._wrapping_algorithm, self.key_name
```
mattsb42-aws: Please make all function calls with more than one argument named parameter calls.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+        :param decryption_materials: Contains verification key, algorithm, encryption context and keyring trace.
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys.
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Contains verification key, algorithm, data_encryption_key, encryption context and keyring trac
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+
+        # Decrypt data key
+        expected_key_info_len = len(self._key_info_prefix) + self._wrapping_algorithm.algorithm.iv_len
```
mattsb42-aws: `encrypted_data_keys` is an iterable of `EncryptedDataKey`. The logic in this method expects a single `EncryptedDataKey`.

Once you address that, you'll need to also add checks to stop iterating through the encrypted data keys once the data encryption key is set.

Ditto on `on_decrypt` in the other keyring.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    :type wrapping_algorithm: WrappingAlgorithm
+    """
+
+    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
+    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(six.binary_type))
+    _wrapping_key = attr.ib(hash=True, repr=False, validator=attr.validators.instance_of(WrappingKey))
+    _wrapping_algorithm = attr.ib(repr=False, validator=attr.validators.instance_of(WrappingAlgorithm))
+
+    def __attrs_post_init__(self):
+
```
mattsb42-aws: Both assignments here need to assign to `self.VARIABLE_NAME`.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    :param bytes key_name: Key ID
+    :param wrapping_key: Encryption key with which to wrap plaintext data key
+    :type wrapping_key: WrappingKey
+    :param wrapping_algorithm: Wrapping Algorithm with which to wrap plaintext data key
+    :type wrapping_algorithm: WrappingAlgorithm
+    :param key_provider: Complete information about the key in the keyring
+    :type key_provider: MasterKeyInfo
+    """
+
+    key_namespace = attr.ib(validator=attr.validators.instance_of(str))
```
mattsb42-aws:
```suggestion
    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
```

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(bytes))
```
mattsb42-aws:
```suggestion
    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(six.binary_types))
```

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    :param str key_namespace: String defining the keyring ID
+    :param bytes key_name: Key ID
+    :param wrapping_key: Encryption key with which to wrap plaintext data key
+    :type wrapping_key: WrappingKey
+    :param wrapping_algorithm: Wrapping Algorithm with which to wrap plaintext data key
+    :type wrapping_algorithm: WrappingAlgorithm
+    """
+
+    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
+    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(six.binary_type))
```
mattsb42-aws: Don't mess with the `hash` settings for any `attr.ib` calls. We set those elsewhere for legacy compatibility reasons, but we should not set them anymore now that we require a newer version of `attrs`.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,221 @@
...
+    :param key_provider: Complete information about the key in the keyring
+    :type key_provider: MasterKeyInfo
+    """
+
+    key_namespace = attr.ib(validator=attr.validators.instance_of(str))
+    key_name = attr.ib(hash=True, validator=attr.validators.instance_of(bytes))
+    _wrapping_key = attr.ib(hash=True, repr=False, validator=attr.validators.instance_of(WrappingKey))
+    _wrapping_algorithm = attr.ib(repr=False, validator=attr.validators.instance_of(WrappingAlgorithm))
+
+    _key_provider = MasterKeyInfo(provider_id=key_namespace, key_info=key_name)
```
mattsb42-aws: This needs to go in a `__attrs_post_init__` method.

MeghaShetty: Made all changes

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+                # EncryptedData to raw key string
+                try:
+                    plaintext_data_key = self._wrapping_key_structure.decrypt(
+                        encrypted_wrapped_data_key=encrypted_wrapped_key,
+                        encryption_context=decryption_materials.encryption_context,
+                    )
+
+                except Exception as error:  # pylint: disable=broad-except
+                    logger = logging.getLogger()
+                    logger.error(error.__class__.__name__, ":", str(error))
```
mattsb42-aws: Since this is a case where if we reach this point, we would reasonably expect to be able to decrypt this data key, I think the right approach here is to log this as an exception[1] with a message that identifies this keyring (ex[2]).

[1] https://docs.python.org/3/library/logging.html#logging.Logger.exception
[2] https://github.com/aws/aws-encryption-sdk-python/blob/35e6549dbd44c72f2c5ff44a53bf51ce9da63195/src/aws_encryption_sdk/key_providers/kms.py#L264-L267

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+        :returns: Optionally modified decryption materials
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        if decryption_materials.data_encryption_key:
+            return decryption_materials
+
+        # Decrypt data key
+        expected_key_info_len = len(self._key_info_prefix) + self._wrapping_algorithm.algorithm.iv_len
+        for key in encrypted_data_keys:
+            if (
```
mattsb42-aws: To improve the readability of this section, let's invert this logic. Rather than entering a deeper nested logic path if the case is true, `continue` if that case is false.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+                        encrypted_wrapped_data_key=encrypted_wrapped_key,
+                        encryption_context=decryption_materials.encryption_context,
+                    )
+
+                except Exception as error:  # pylint: disable=broad-except
+                    logger = logging.getLogger()
+                    logger.error(error.__class__.__name__, ":", str(error))
+                    return decryption_materials
+
+                if plaintext_data_key:
```
mattsb42-aws: What logic path will result in this being false?
Should that be an error instead?

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
+    key_name = attr.ib(validator=attr.validators.instance_of(six.binary_type))
+    # _wrapping_key = attr.ib(repr=False, validator=attr.validators.in_([rsa.RSAPrivateKey, rsa.RSAPublicKey]))
+    # ----- need to figure out how to do this correctly
+    _wrapping_key = attr.ib(repr=False, validator=attr.validators.instance_of(object))
+    _wrapping_algorithm = attr.ib(repr=False, validator=attr.validators.instance_of(WrappingAlgorithm))
+
+    @classmethod
+    def fromPEMEncoding(cls):
+        key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
```
mattsb42-aws: Unfortunately, attrs doesn't work like this on classmethods. We'll need to take inputs like any other method.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+        key_name = attr.ib(validator=attr.validators.instance_of(six.binary_type))
+        _encoded_key = attr.ib(validator=attr.validators.instance_of(six.binary_type))
+        _password = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(six.binary_type)))
+        _wrapping_algorithm = attr.ib(repr=False, validator=attr.validators.instance_of(WrappingAlgorithm))
+        if _password:
```
mattsb42-aws: Password is not a differentiator between private and public, it is an optional parameter that might be needed to load an encrypted private key.

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.load_pem_private_key

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+    :type wrapping_algorithm: WrappingAlgorithm
+    :param key_provider: Complete information about the key in the keyring
+    :type key_provider: MasterKeyInfo
+    """
+
+    key_namespace = attr.ib(validator=attr.validators.instance_of(six.string_types))
+    key_name = attr.ib(validator=attr.validators.instance_of(six.binary_type))
+    # _wrapping_key = attr.ib(repr=False, validator=attr.validators.in_([rsa.RSAPrivateKey, rsa.RSAPublicKey]))
+    # ----- need to figure out how to do this correctly
+    _wrapping_key = attr.ib(repr=False, validator=attr.validators.instance_of(object))
```
mattsb42-aws: I've been thinking about this, and I think it might be best to expand this to two optional parameters, of which at least one must be provided: a private key and a public key.

Then, in post-init we can load the public key from the private key if only the private key was provided.

This makes the API more explicit and hopefully simpler to use, and also makes our internal logic simpler because we don't have to keep checking to see what we have.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+        # EncryptedData to EncryptedDataKey
+        encrypted_data_key = serialize_wrapped_key(
+            key_provider=self._key_provider,
+            wrapping_algorithm=self._wrapping_algorithm,
+            wrapping_key_id=self.key_name,
+            encrypted_wrapped_key=encrypted_wrapped_key,
+        )
+
+        # Update Keyring Trace
+        if encrypted_data_key:
```
mattsb42-aws: What is the logic flow that would result in this being false but also not have thrown an error?

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+        """Attempt to decrypt the encrypted data keys.
+
+        :param decryption_materials: Decryption materials for the keyring to modify.
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys.
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        if isinstance(self._wrapping_key, rsa.RSAPrivateKey):
```
mattsb42-aws: Ditto on improving readability by inverting this check.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+            # Decrypt data key
+            for key in encrypted_data_keys:
+                if key.key_provider == self._key_provider:
```
mattsb42-aws: Ditto on improving readability by inverting this check.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+                        wrapping_key_id=self.key_name,
+                        wrapped_encrypted_key=key,
+                    )
+                    try:
+                        plaintext_data_key = self._wrapping_key.decrypt(
+                            ciphertext=encrypted_wrapped_key.ciphertext, padding=self._wrapping_algorithm.padding
+                        )
+                    except Exception as error:  # pylint: disable=broad-except
+                        logger = logging.getLogger()
+                        logger.error(error.__class__.__name__, ":", str(error))
```
mattsb42-aws: See comments on analogous portion of `RawAESKeyring`.

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
...
+                        return decryption_materials
+
+                    if plaintext_data_key:
```
mattsb42-aws: What is the logic flow where we have arrived at this line but this check is false?

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,390 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
MeghaShetty: I think "from aws_encryption_sdk.keyring.raw.aes import RawAesKeyring" is more readable as it is detailed.

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
```
mattsb42-aws: We also need to have unit tests for the `generate_data_key`:


- [ ] raise GenerateKeyError if os.urandom fails
- [ ] input MUST NOT contain data encryption key
- [ ] output MUST contain data encryption key and EXACTLY ONE WRAPPING_KEY_GENERATED_DATA_KEY trace for the input keyring

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,79 @@
...
+            "\xa1\x08t6{\xf2\xfd\xf1\xb3\x02\x01\x10\x80;D\xa4\xed`qP~c\x0f\xa0d"
+            "\xd5\xa2Kj\xc7\xb2\xc6\x1e\xec\xfb\x0fK\xb2*\xd5\t2\x81pR\xee\xd1"
+            '\x1a\xde<"\x1b\x98\x88\x8b\xf4&\xdaB\x95I\xd2\xff\x10\x13\xfc\x1aX'
+            "\x08,/\x8b\x8b"
+        ),
+    )
+]
+
+
+def test_keyring_no_encrypt():
```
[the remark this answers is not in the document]
MeghaShetty: I made all changes to the base API and the tests and made a new pull request for them

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+
+    original_number_of_encrypted_data_keys = len(
+        get_encryption_materials_without_data_encryption_key().encrypted_data_keys
+    )
+
+    test = test_raw_aes_keyring.on_encrypt(encryption_materials=get_encryption_materials_without_data_encryption_key())
+
+    # Check if data key is generated
+    assert test.data_encryption_key and test.data_encryption_key is not None
```
mattsb42-aws: This value being truthy *but not `None`* doesn't really tell us anything. There is no valid falsy value for it other than `None`, and I worry that putting a check like this might lead a future reader to think that there is.
```suggestion
    assert test.data_encryption_key is not None
```

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+    test_raw_aes_keyring = RawAESKeyring(
+        key_namespace=_PROVIDER_ID,
+        key_name=_KEY_ID,
+        wrapping_algorithm=WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING,
+        wrapping_key=_WRAPPING_KEY,
+    )
+    test_raw_aes_keyring.on_decrypt(
+        decryption_materials=get_decryption_materials_with_data_encryption_key(),
+        encrypted_data_keys=[
+            EncryptedDataKey(
```
mattsb42-aws: Let's move this construction up to a constant.

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+@pytest.fixture
+def patch_generate_data_key(mocker):
+    mocker.patch.object(aws_encryption_sdk.keyring.raw_keyring, "generate_data_key")
+    return aws_encryption_sdk.keyring.raw_keyring.generate_data_key
+
+
+@pytest.fixture
+def patch_decrypt_data_key(mocker):
```
mattsb42-aws: Please add something to this fixture name to clarify that is is patching decrypt *on WrappingKey*.

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+            encrypted_flag_count += 1
+
+    assert generated_flag_count == 1
+
+    assert len(test.encrypted_data_keys) == original_number_of_encrypted_data_keys + 1
+
+    assert encrypted_flag_count == 1
+
+
+def test_on_decrypt_when_data_key_given(patch_decrypt_data_key):
```
mattsb42-aws: Let's parameterize this for an additional decryption materials case:
- [x] data key in input materials and EDKs provided
- [ ] data key in input materials and *no* EDKs provided

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+            )
+        ],
+    )
+    for keyring_trace in test.keyring_trace:
+        if keyring_trace.wrapping_key.key_info == _KEY_ID:
+            # Check keyring trace does not contain KeyringTraceFlag.WRAPPING_KEY_DECRYPTED_DATA_KEY
+            assert KeyringTraceFlag.WRAPPING_KEY_DECRYPTED_DATA_KEY not in keyring_trace.flags
+
+
+def test_on_decrypt_when_data_key_and_edk_not_provided(patch_decrypt_data_key):
```
mattsb42-aws: Let's combine this and `test_on_decrypt_when_data_key_not_provided_and_edk_not_in_keyring` into a single test with two parameterized values for `encrypted_data_keys`.

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+        key_name=_KEY_ID,
+        wrapping_algorithm=WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING,
+        wrapping_key=_WRAPPING_KEY,
+    )
+
+    test_raw_aes_keyring.on_decrypt(
+        decryption_materials=get_decryption_materials_without_data_encryption_key(),
+        encrypted_data_keys=[_ENCRYPTED_DATA_KEY_AES],
+    )
+    assert patch_decrypt_data_key.called_once_with(
```
mattsb42-aws: `Mock.called_once_with()` doesn't exist. Use `Mock.assert_called_once_with()`.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,343 @@
...
+        signing_key=_SIGNING_KEY,
+    )
+    generate_data_key(
+        encryption_materials=encryption_materials_without_data_key,
+        key_provider=MasterKeyInfo(provider_id=_PROVIDER_ID, key_info=_KEY_ID),
+    )
+
+    generate_flag_count = 0
+    for keyring_trace in encryption_materials_without_data_key.keyring_trace:
+        if KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY in keyring_trace.flags:
```
mattsb42-aws: Let's also make sure that the key provider is as expected too.

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@pytest.fixture
+def patch_generate_data_key(mocker):
+    mocker.patch.object(aws_encryption_sdk.keyring.raw_keyring, "generate_data_key")
+    return aws_encryption_sdk.keyring.raw_keyring.generate_data_key
+
+
+@pytest.fixture
+def patch_decrypt_data_key(mocker):
```
mattsb42-aws: Ditto on naming: make it clear if I'm just looking at a test where this fixture is being consumed, what this is patching.

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,311 @@
...
+            public_exponent=_PUBLIC_EXPONENT, key_size=_KEY_SIZE, backend=_BACKEND
+        ),
+    )
+    assert test.key_namespace == _PROVIDER_ID
+    assert test.key_name == _KEY_ID
+    assert test._wrapping_algorithm == WrappingAlgorithm.RSA_OAEP_SHA256_MGF1
+    assert isinstance(test._private_wrapping_key, rsa.RSAPrivateKey)
+
+
+def test_missing_required_parameters():
```
mattsb42-aws: See comments on missing/invalid parameters in raw AES keyring tests.

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,311 @@
...
+            # Check keyring trace does not contain KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY
+            assert KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY not in keyring_trace.flags
+
+
+def test_on_encrypt_when_data_encryption_key_not_given():
+    test_raw_rsa_keyring = RawRSAKeyring(
+        key_namespace=_PROVIDER_ID,
+        key_name=_KEY_ID,
+        wrapping_algorithm=WrappingAlgorithm.RSA_OAEP_SHA256_MGF1,
+        private_wrapping_key=rsa.generate_private_key(
```
mattsb42-aws: We have some valid RSA keys over in `test_values`. Let's just use one of those rather than generating a new one every time.

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+    original_number_of_encrypted_data_keys = len(
+        get_encryption_materials_without_data_encryption_key().encrypted_data_keys
+    )
+
+    test = test_raw_rsa_keyring.on_encrypt(encryption_materials=get_encryption_materials_without_data_encryption_key())
+
+    # Check if data key is generated
+    assert test.data_encryption_key and test.data_encryption_key is not None
```
mattsb42-aws: Ditto on falsy values.
```suggestion
    assert test.data_encryption_key is not None
```

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+    test = test_raw_rsa_keyring.on_decrypt(
+        decryption_materials=get_decryption_materials_without_data_encryption_key(),
+        encrypted_data_keys=test_raw_rsa_keyring.on_encrypt(
+            encryption_materials=get_encryption_materials_without_data_encryption_key()
+        ).encrypted_data_keys,
+    )
+    decrypted_flag_count = 0
+
```
mattsb42-aws: Let's also go ahead and throw in an assert that the data encryption key is set.

test/unit/test_keyring_raw_aes.py
```diff
@@ -0,0 +1,276 @@
...
+    test = raw_aes_keyring
+    assert test.key_name == _KEY_ID
+    assert test.key_namespace == _PROVIDER_ID
+    assert test._wrapping_algorithm == WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING
+    assert test._wrapping_key == _WRAPPING_KEY
+
+
+@pytest.mark.parametrize(
+    "key_namespace, key_name, wrapping_algorithm, wrapping_key",
+    (
```
mattsb42-aws: The test cases I'd like to see here are:

- [ ] for each required parameter, that parameter's value is `None` but all other parameters are valid
- [ ] for each parameter with a validator, that parameter's value is invalid but all other parameters are valid (for required parameters there can be a lot of overlap between this and the previous where the same test case can address both)

For example, if I had a class that accepted parameters A, B, and C, of which A and B are required `int`s and C is an optional string, I would want:

```python
# A, B, C
None, 42, "petunias"
42, None, "petunias"
42, 42, 42
```

test/unit/test_keyring_raw_rsa.py
```diff
@@ -0,0 +1,243 @@
...
+    test = raw_rsa_keyring
+    assert test.key_namespace == _PROVIDER_ID
+    assert test.key_name == _KEY_ID
+    assert test._wrapping_algorithm == WrappingAlgorithm.RSA_OAEP_SHA256_MGF1
+    assert isinstance(test._private_wrapping_key, rsa.RSAPrivateKey)
+
+
+@pytest.mark.parametrize(
+    "key_namespace, key_name, wrapping_algorithm, private_wrapping_key, public_wrapping_key",
+    (
```
mattsb42-aws: See requested test cases in raw AES keyring tests.

PR #166, opened 2019-07-01
Multi keyrings

MeghaShetty:
*Issue #, if available:*

*Description of changes:*
Dependent on the merging of PRs for base keyring API, keyring materials and raw keyrings.


src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+class MultiKeyring(Keyring):
+    """Public class for Multi Keyring.
+
+    :param generator: Generator keyring used to generate data encryption key
+    :type generator: Keyring
+    :param list children: List of keyrings used to encrypt the data encryption key
+    :raises EncryptKeyError: if encryption of data key fails for any reason
+    """
+
+    generator = attr.ib(validator=attr.validators.instance_of(Keyring))
```
mattsb42-aws: Both generator and children need to be optional, with validation in `__attrs_post_init__` that at least one of them is set. You can see a similar pattern here: https://github.com/aws/aws-encryption-sdk-python/blob/keyring/src/aws_encryption_sdk/streaming_client.py#L99-L105

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+    children = attr.ib(validator=attr.validators.instance_of(list))
```
mattsb42-aws: This validator needs to be `deep_iterable`[1] with a `member_validator` checking that all members are `Keyring`s.

[1] https://attrs.readthedocs.io/en/stable/api.html#attr.validators.deep_iterable

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+        and encrypt it using any available wrapping key in any child keyring.
+
+        :param encryption_materials: Encryption materials for keyring to modify.
+        :type encryption_materials: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :returns: Optionally modified encryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :raises EncryptKeyError: if unable to encrypt data key.
+        """
+        # Check if generator keyring is not provided and data key is not generated
+        if not self.generator and not encryption_materials.data_encryption_key:
```
mattsb42-aws: Per my above comment, if `self.generator` is not provided by the caller, its value will be `None`. Anywhere we need to determine if the generator was set, use the `is` operator rather than just checking for truthy/falsy.

http://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+            raise EncryptKeyError("Generator keyring not provided.")
```
mattsb42-aws:
```suggestion
            raise EncryptKeyError("Generator keyring not provided and encryption materials do not already contain a plaintext data key.")
```

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Check if generator keyring is provided and data key is generated
```
mattsb42-aws: This is actually ok, not a failure mode.

This is something I was misremembering when we discussed this before.

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        # Check if plaintext data key exists
+        if decryption_materials.data_key:
+            return decryption_materials
+
+        # Call on_decrypt on all keyrings till decryption is successful
+        for keyring in self.children:
```
mattsb42-aws: We also need to include the generator in the decrypt path.

Also, make sure that we do not use the generator if it is `None`.

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+
+        :param decryption_materials: Decryption materials for keyring to modify.
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys.
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        # Check if plaintext data key exists
+        if decryption_materials.data_key:
```
mattsb42-aws: Please use the new API for consistency across encryption/decryption materials.
```suggestion
        if decryption_materials.data_encryption_key:
```

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,92 @@
...
+        # Check if generator keyring is not provided and data key is not generated
+        if not self.generator and not encryption_materials.data_encryption_key:
+            raise EncryptKeyError("Generator keyring not provided.")
+
+        # Check if generator keyring is provided and data key is generated
+        if self.generator and encryption_materials.data_encryption_key:
+            raise EncryptKeyError("Data encryption key already exists.")
+
+        # Call on_encrypt on the generator keyring
+        encryption_materials = self.generator.on_encrypt(encryption_materials)
```
mattsb42-aws: What if the generator was not set?

MeghaShetty: Made all changes

test/functional/test_f_multi_keyring.py
```diff
@@ -0,0 +1,150 @@
...
+                key_namespace=_PROVIDER_ID,
+                key_name=_KEY_ID,
+                wrapping_algorithm=WrappingAlgorithm.AES_128_GCM_IV12_TAG16_NO_PADDING,
+                wrapping_key=_WRAPPING_KEY_AES,
+            ),
+        ]
+    ),
+]
+
+
```
mattsb42-aws: Additional behaviors we need to test:

## constructor
- [ ] constructor called with a generator but no children must succeed
- [ ] constructor called with children but no generator must succeed
- [ ] constructor called a generator and no children must throw `TypeError`
- [ ] constructor called with a generator or any children that are not `Keyring`s must throw `TypeError`

## on_encrypt
- [ ] given no generator, `on_encrypt` with no `data_encryption_key` must throw `EncryptKeyError`
- [ ] given a keyring that does nothing (make one of these as a test util) as the generator, `on_encrypt` with no `data_encryption_key` must throw `EncryptKeyError` (or `GenerateKeyError` if you change to that)
- [ ] given a keyring that always adds an EDK (make one of these as a test util) in each of the generator and multiple children, the `on_encrypt` return value must contain `len(children) + (1 if generator else 0)` more encrypted data keys than the input materials (test this for each of: generator but no children, children but no generator, both children and generator)
- [ ] given a keyring that generates but does not add an EDK, and no children, the `on_encrypt` return value must contain exactly the number of EDKs as the input value (test with both no input EDKs and >0)
- [ ] given input materials that already contain a `data_encryption_key`, every keyring must be called (you can use mock keyrings for this to simplify checking)

## on_decrypt
- [ ] given decryption materials that already contain a `data_encryption_key`, we must not call any of the keyrings (you can use mock keyrings for this to simplify checking)
- [ ] given a keyring that never adds an EDK in each of the generator and mutliple children, every keyring must be called (you can use mock keyrings for this to simplify checking)
- [ ] given decryption materials that do *not* already contain a `data_encryption_key`, once a keyring that adds an EDK is encountered, we must not call any further keyrings (you can use mock keyrings for this to simplify checking)

One good starting point when thinking about what to test is to step through your code, and any time you encounter a different behavior, make sure that you are testing that. One metric to look at to start measuring what you have and have not tested is to look at the coverage report and look for lines *and logic branches* that are not covered.

Coverage is not a perfect metric, but it's a good starting point and a reasonably good negative metric: ie, good coverage numbers don't mean that you have complete tests, but bad coverage numbers do mean you don't have complete tests.

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,107 @@
...
+        and encrypt it using any available wrapping key in any child keyring.
+
+        :param encryption_materials: Encryption materials for keyring to modify.
+        :type encryption_materials: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :returns: Optionally modified encryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :raises EncryptKeyError: if unable to encrypt data key.
+        """
+        # Check if generator keyring is not provided and data key is not generated
+        if self.generator is None and not encryption_materials.data_encryption_key:
```
mattsb42-aws: I missed this before. The unset value for `CryptographicMaterials.data_encryption_key` is `None`[1]. We should make sure that we are checking for `None` correctly everywhere we check it.

```suggestion
        if self.generator is None and encryption_materials.data_encryption_key is None:
```

[1] https://github.com/aws/aws-encryption-sdk-python/blob/keyring/src/aws_encryption_sdk/materials_managers/__init__.py#L91-L93

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,107 @@
...
+        # type: (DecryptionMaterials, Iterable[EncryptedDataKey]) -> DecryptionMaterials
+        """Attempt to decrypt the encrypted data keys.
+
+        :param decryption_materials: Decryption materials for keyring to modify.
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys.
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
```
mattsb42-aws: I've been trying to think of a way that we could simplify this function. It occurred to me today as I was reading through this that if we swap the order of operations in the loop, we can get rid of the initial check.

ie:
```python
        for keyring in self._decryption_keyrings:
            if decryption_materials.data_encryption_key:
                return decryption_materials
            decryption_materials = keyring.on_decrypt(decryption_materials, encrypted_data_keys)

        return decryption_materials
```

This would still handle each case correctly, and would be slightly fewer lines of code to maintain.
However, I'm not sure whether or not this would be an overly clever premature optimization.
I'm not opposed to keeping the extra initial check if it aids the readability.

@MeghaShetty I'm interested in your perspective on this. Which do you think is more readable?

MeghaShetty: This code is readable as well. It is easy to understand. I like this modification! I'll make the changes.

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,103 @@
...
+                "Generator keyring not provided "
+                "and encryption materials do not already contain a plaintext data key."
+            )
+
+        # Call on_encrypt on the generator keyring if it is provided
+        if self.generator is not None:
+            encryption_materials = self.generator.on_encrypt(encryption_materials)
+
+        # Check if data key is generated
+        if not encryption_materials.data_encryption_key:
```
mattsb42-aws:
```suggestion
        if encryption_materials.data_encryption_key is None:
```

src/aws_encryption_sdk/keyring/multi_keyring.py
```diff
@@ -0,0 +1,103 @@
...
+        :param decryption_materials: Decryption materials for keyring to modify.
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys.
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials.
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        # Call on_decrypt on all keyrings till decryption is successful
+        for keyring in self._decryption_keyrings:
+            if decryption_materials.data_encryption_key:
```
mattsb42-aws:
```suggestion
            if decryption_materials.data_encryption_key is not None:
```

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,353 @@
...
+    """Helper function for the on_encrypt function of keyring.
+
+    :param encryption_materials: Encryption materials for the keyring to modify.
+    :type encryption_materials: aws_encryption_sdk.materials_managers.EncryptionMaterials
+    :param key_provider: Information about the key in the keyring.
+    :type key_provider: MasterKeyInfo
+    :return bytes: Plaintext data key
+    """
+    # Check if data key already exists
+    if not encryption_materials.data_encryption_key:
```
mattsb42-aws:
```suggestion
    if encryption_materials.data_encryption_key is None:
```

src/aws_encryption_sdk/keyring/raw_keyring.py
```diff
@@ -0,0 +1,353 @@
...
+        """Attempt to decrypt the encrypted data keys.
+
+        :param decryption_materials: Decryption materials for the keyring to modify
+        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :param encrypted_data_keys: List of encrypted data keys
+        :type: List of `aws_encryption_sdk.structures.EncryptedDataKey`
+        :returns: Optionally modified decryption materials
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        """
+        if decryption_materials.data_encryption_key:
```
mattsb42-aws:
```suggestion
        if decryption_materials.data_encryption_key is not None:
```

test/functional/test_f_multi_keyring.py
```diff
@@ -0,0 +1,150 @@
...
+    decryption_materials = DecryptionMaterials(
+        algorithm=Algorithm.AES_256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384, verification_key=b"ex_verification_key"
+    )
+
+    # Call on_decrypt function for the keyring
+    decryption_materials = multi_keyring.on_decrypt(
+        decryption_materials=decryption_materials, encrypted_data_keys=encryption_materials.encrypted_data_keys
+    )
+
+    if decryption_materials.data_encryption_key:
```
mattsb42-aws: This cycle test should carry an expectation that this is *always* true.

If there are scenarios to tests where this is *not* true, that should be tested in a different test function.

test/unit/test_keyring_multi.py
```diff
@@ -0,0 +1,242 @@
...
+
+    test = test_multi_keyring.on_encrypt(encryption_materials=get_encryption_materials_with_encrypted_data_key())
+    assert len(test.encrypted_data_keys) == len(get_encryption_materials_with_encrypted_data_key().encrypted_data_keys)
+
+
+def test_on_decrypt_when_data_encryption_key_given():
+    test_multi_keyring = MultiKeyring(generator=mock_generator, children=[mock_child_1, mock_child_2])
+    test_multi_keyring.on_decrypt(decryption_materials=get_decryption_materials_with_data_key(), encrypted_data_keys=[])
+    for keyring in test_multi_keyring._decryption_keyrings:
+        keyring.on_decrypt.assert_not_called()
```
mattsb42-aws: This is a problematic corner of Python's `mock` history...

Always use `assert not X.called`, not `X.assert_not_called()`.

The `assert_not_called` method was added in Python 3.5 because people kept assuming that it should be there; in earlier versions of Python 3 and in the `mock` library for Python 2, this will either silently work (because the `Mock.assert_not_called` is a mock and therefore returns a mock) or will throw an error, depending on the version of the `mock` library.

```suggestion
        assert not keyring.on_decrypt.called
```

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called

test/unit/test_keyring_multi.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def test_on_decrypt_every_keyring_called_when_data_encryption_key_not_added():
+    mock_generator.on_decrypt.return_value = get_decryption_materials_without_data_key()
```
mattsb42-aws: Let's be a bit less brittle and decouple these from what we assume the input will be:
```suggestion
    mock_generator.on_decrypt.side_effect = lambda x: x
```

This changes the call behavior of `mock_generator.on_decrypt` to be the identity function, but still retains the mock instrumenting.

test/unit/test_keyring_multi.py
```diff
@@ -0,0 +1,242 @@
...
+#         children=[
+#             mock_child_3,
+#             mock_child_1,
+#             mock_child_2,
+#         ]
+#     )
+#     test_multi_keyring.on_decrypt(decryption_materials=get_decryption_materials_without_data_key(),
+#                                   encrypted_data_keys=[])
+#     mock_generator.on_decrypt.assert_called()
+#     mock_child_3.on_decrypt.assert_called()
```
mattsb42-aws: Similar to the issue with `assert_not_called`, this was only added in Python 3.6 :(

Use `assert X.called` instead.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called

test/unit/test_keyring_multi.py
```diff
@@ -0,0 +1,242 @@
...
+        MultiKeyring(children=[WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING])
+    assert exc_info.errisinstance(TypeError)
+
+
+def test_on_encrypt_with_no_generator_no_data_encryption_key():
+    test_multi_keyring = get_multi_keyring_with_no_generator()
+    with pytest.raises(EncryptKeyError) as exc_info:
+        test_multi_keyring.on_encrypt(encryption_materials=get_encryption_materials_without_data_key())
+    assert exc_info.match(
+        "Generator keyring not provided and encryption materials do not already " "contain a plaintext data key."
```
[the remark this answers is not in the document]
MeghaShetty: Yes. Changed it and also all other changes.

test/unit/test_keyring_multi.py
```diff
@@ -51,28 +45,44 @@
...
+
+
+@pytest.fixture
+def keyring_which_only_generates():
+    return OnlyGenerateKeyring()
+
+
+@pytest.fixture
+def mock_generator():
+    mock_generator = MagicMock()
```
mattsb42-aws: As a general good practice, we should avoid overriding a name from a higher scope (in this case, the function name).

test/unit/test_keyring_multi.py
```diff
@@ -94,7 +104,12 @@ def test_keyring_with_generator_but_no_children():
             wrapping_algorithm=WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING,
         )
     )
-    assert isinstance(test_multi_keyring.generator, RawAESKeyring)
+    assert isinstance(test_multi_keyring.generator, Keyring)
```
mattsb42-aws: What I meant here is that what we are interested in is that the *specific instance* of this keyring was assigned to the `generator` attribute, not just what class the variable assigned to the `generator` attribute is.

ex of what I meant to suggest:
```python
generator_keyring = RawAESKeyring(...)
test_multi_keyring = MultiKeyring(generator=generator, ...)

assert test_multi_keyring.generator is generator_keyring
```

The below lines sort of do this, but they would also pass if we made a copy of `generator_keyring` and used that instead.

test/unit/test_keyring_multi.py
```diff
@@ -94,7 +104,12 @@ def test_keyring_with_generator_but_no_children():
...
+    assert test_multi_keyring.generator.key_namespace == _PROVIDER_ID
+    assert test_multi_keyring.generator.key_name == _KEY_ID
+    assert test_multi_keyring.generator._wrapping_key == _WRAPPING_KEY_AES
+    assert test_multi_keyring.generator._wrapping_algorithm == WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING
+    assert test_multi_keyring.children == ()
```
mattsb42-aws: Hedging our bets just in case we for some reason decide to use some other container class in the future.

```python
assert not test_multi_keyring.children
```

test/unit/test_keyring_multi.py
```diff
@@ -108,7 +123,12 @@ def test_keyring_with_children_but_no_generator():
             )
         ]
     )
-    assert isinstance(test_multi_keyring.children, list)
```
mattsb42-aws: See comments on previous test re matching attributes of keyring vs matching identity.

test/unit/test_keyring_multi.py
```diff
@@ -117,23 +137,39 @@ def test_keyring_with_no_generator_no_children():
...
-    assert exc_info.errisinstance(TypeError)
+        MultiKeyring(generator=generator, children=children)
+    assert exc_info.match("('children'|'generator') must be <class 'aws_encryption_sdk.keyring.base.Keyring'>.*")
+
+
+def test_decryption_keyrings():
+    test_multi_keyring = get_multi_keyring_with_generator_and_children()
+    assert test_multi_keyring.generator in test_multi_keyring._decryption_keyrings
+    for child_keyring in test_multi_keyring.children:
+        assert child_keyring in test_multi_keyring._decryption_keyrings
```
mattsb42-aws: Let's also add an assertion that there are no additional keyrings in `_decryption_keyrings`

```python
assert len(test_multi_keyring._decryption_keyrings) == len(test_multi_keyring.children) + 1
```

PR #176, opened 2019-07-23
Base keyrings

MeghaShetty:
*Issue #, if available:*

*Description of changes:*
Made suggested changes in base API and tests

# Check any applicable:


test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,73 @@
...
+            "\xd5\xa2Kj\xc7\xb2\xc6\x1e\xec\xfb\x0fK\xb2*\xd5\t2\x81pR\xee\xd1"
+            '\x1a\xde<"\x1b\x98\x88\x8b\xf4&\xdaB\x95I\xd2\xff\x10\x13\xfc\x1aX'
+            "\x08,/\x8b\x8b"
+        ),
+    )
+]
+
+
+def test_keyring_no_encrypt():
+    class KeyringNoEncrypt(Keyring):
```
mattsb42-aws: `Keyring` will do what you need here just fine; no need to make a special class for it.

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,73 @@
...
+        def on_decrypt(self, _decryption_materials, _encrypted_data_keys):
+            return
```
mattsb42-aws: This is not a valid return value for a keyring `on_encrypt` method.

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert pytest.raises(NotImplementedError)
```
mattsb42-aws: You're not actually testing anything here. `pytest.raises` returns a truthy value, so this will pass, but it's not actually testing anything.

What you want is to use the context manager that `pytest.raises` returns to catch a thrown exception (in this case when you call `Keyring().on_decrypt(...)` and verify that the exception caught it what you expected.

ex:
https://github.com/aws/aws-encryption-sdk-python/blob/4857df9e529601f319d3a04fc1542fbb88160208/test/unit/test_encryption_context.py#L39-L44

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_keyring_no_decrypt():
```
mattsb42-aws: See `test_keyring_no_encrypt` comments.

test/unit/test_utils.py
```diff
@@ -21,11 +21,19 @@
...
 from aws_encryption_sdk.internal.defaults import MAX_FRAME_SIZE, MESSAGE_ID_LENGTH
+from aws_encryption_sdk.materials_managers import EncryptionMaterials, DecryptionMaterials, EncryptedDataKey
 from aws_encryption_sdk.structures import DataKey, EncryptedDataKey, MasterKeyInfo, RawDataKey
+from aws_encryption_sdk.keyring.raw_keyring import RawRSAKeyring, RawAESKeyring
 
 from .test_values import VALUES
 from .unit_test_utils import assert_prepped_stream_identity
 
+try:  # Python 3.5.0 and 3.5.1 have incompatible typing modules
+    from typing import Iterable, Union  # noqa pylint: disable=unused-import
```
mattsb42-aws: Is `Union` actually used anywhere in this module?

test/unit/test_utils.py
```diff
@@ -42,6 +50,24 @@ def test_prep_stream_data_wrap(source):
...
+    def on_encrypt(self, encryption_materials):
+        # type: (EncryptionMaterials) -> EncryptionMaterials
+        return encryption_materials
+
+    def on_decrypt(self, decryption_materials, encrypted_data_keys):
+        # type: (DecryptionMaterials, Iterable[EncryptedDataKey]) -> DecryptionMaterials
+        return decryption_materials
+
+
+class NullRawAESKeyring(RawAESKeyring):
```
mattsb42-aws: The methods on this class do not implement valid Keyring APIs.

test/unit/test_utils.py
```diff
@@ -42,6 +50,24 @@ def test_prep_stream_data_wrap(source):
     assert_prepped_stream_identity(test, io.BytesIO)
 
 
+class NullRawRSAKeyring(RawRSAKeyring):
```
mattsb42-aws: Two thoughts:
1. `IdentityKeyring` would probably be a better name for this: more representative of what it does.
2. Is there any reason you are subclassing from `RawRSAKeyring` rather than `Keyring`?

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,73 @@
...
+            '\x1a\xde<"\x1b\x98\x88\x8b\xf4&\xdaB\x95I\xd2\xff\x10\x13\xfc\x1aX'
+            "\x08,/\x8b\x8b"
+        ),
+    )
+]
+
+
+def test_keyring_no_encrypt():
+    class KeyringNoEncrypt(Keyring):
+        def on_decrypt(self, _decryption_materials, _encrypted_data_keys):
```
mattsb42-aws: NP: You could equally well pass in an empty list for `encrypted_data_keys` and get the same behavior for this test.

test/unit/test_utils.py
MeghaShetty: I added that while writing a test for raw RSA Keyrings and I forgot to delete it later. I will delete that.

test/unit/test_utils.py
MeghaShetty: The null keyrings were added when I was writing unit tests. I didn't use them later. I will delete them.

test/unit/test_keyring_base.py
```diff
@@ -0,0 +1,51 @@
...
+"""Unit tests for base keyring."""
+
+import pytest
+
+from aws_encryption_sdk.identifiers import Algorithm
+from aws_encryption_sdk.keyring.base import Keyring
+from aws_encryption_sdk.materials_managers import DecryptionMaterials, EncryptionMaterials
+
+try:  # Python 3.5.0 and 3.5.1 have incompatible typing modules
+    from typing import Iterable  # noqa pylint: disable=unused-import
```
mattsb42-aws: It looks like we're not using this anymore.

test/unit/test_utils.py
```diff
@@ -26,6 +26,12 @@
 from .test_values import VALUES
 from .unit_test_utils import assert_prepped_stream_identity
 
+try:  # Python 3.5.0 and 3.5.1 have incompatible typing modules
+    from typing import Iterable  # noqa pylint: disable=unused-import
```
mattsb42-aws: It looks like we're not using this anymore.

PR #177, opened 2019-07-24
Example for using multiple keys in multiple regions

caitlin-tibbetts:
Issue #156, example A2.B1.C1

Added an example and a test for when multiple keys in multiple regions are used to encrypt and decrypt a passed-in message.


mattsb42-aws:
Please run:
```shell
tox -re autoformat
tox -re {flake8,pylint}-examples
```

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,48 @@
...
+
+def encrypt_decrypt(key_arn1, key_arn2, region_name1, region_name2, source_plaintext, botocore_session=None):
+    """Encrypts and then decrypts a string under one KMS customer master key (CMK).
+
+    :param str key_arn: Amazon Resource Name (ARN) of the KMS CMK
+    :param bytes source_plaintext: Data to encrypt
+    :param botocore_session: existing botocore session instance
+    :type botocore_session: botocore.session.Session
+    """
+    kwargs = dict(key_ids=[key_arn1, key_arn2], region_names=[region_name1, region_name2])
```
mattsb42-aws: Since we have stated that these need to be Arns, I would prefer to leave off  the `region_names` parameter. It is less useful that I initially thought it would be, and IME tends to confuse the reader more than it helps.

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,48 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Example showing basic encryption and decryption of a value already in memory using multiple KMS CMKs in multiple regions."""
+import aws_encryption_sdk
+
+
+def encrypt_decrypt(key_arn1, key_arn2, region_name1, region_name2, source_plaintext, botocore_session=None):
+    """Encrypts and then decrypts a string under one KMS customer master key (CMK).
+
```
mattsb42-aws: Copied docstring is no longer accurate.

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,48 @@
...
+    :param bytes source_plaintext: Data to encrypt
+    :param botocore_session: existing botocore session instance
+    :type botocore_session: botocore.session.Session
+    """
+    kwargs = dict(key_ids=[key_arn1, key_arn2], region_names=[region_name1, region_name2])
+
+    if botocore_session is not None:
+        kwargs["botocore_session"] = botocore_session
+
+    # Create master key provider using the ARN of the key and the session (botocore_session)
```
mattsb42-aws: Copied comment is no longer accurate.

tox.ini
```diff
@@ -55,7 +55,7 @@ commands =
     examples: {[testenv:base-command]commands} examples/test/ -m examples
     all: {[testenv:base-command]commands} test/ examples/test/
     manual: {[testenv:base-command]commands}
-
+
```
mattsb42-aws: Did your IDE do this?
I'd prefer not to have a line of extra whitespace like this.

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,48 @@
...
+    kms_key_provider = aws_encryption_sdk.KMSMasterKeyProvider(**kwargs)
+
+    # Encrypt the plaintext using the AWS Encryption SDK. It returns the encrypted message and the header
+    ciphertext, encrypted_message_header = aws_encryption_sdk.encrypt(
+        source=source_plaintext, key_provider=kms_key_provider
+    )
+
```
mattsb42-aws: To demonstrate what just happened, let's add an assert that there are two encrypted data keys in `encrypted_message_header`.

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Example showing basic encryption and decryption of a value already in memory using multiple KMS CMKs in multiple regions."""
+import aws_encryption_sdk
+
+
+def encrypt_decrypt(key_arn1, key_arn2, region_name1, region_name2, source_plaintext, botocore_session=None):
```
ttjsu-aws: I would the encrypt and decrypt functionality to two different functions, so that it is more readable and can be re-used for encryption and decryption under the different keys, if needed. 

Refer:  https://github.com/aws/aws-encryption-sdk-c/blob/master/examples/kms_discovery.cpp to give you an idea on what are are looking for in this example.

examples/src/multiple_kms_cmk_regions.py
ttjsu-aws: What we are looking for in this example: Encrypt a string using two different KMS CMKs in two different regions, for ex:  one in `us-west-2` and one in `eu-central-1`. This should have the effect of encrypting the data key separately with each CMK. Then show that we can decrypt the data with either one of the keys.

examples/test/test_i_multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,31 @@
...
+
+from ..src.multiple_kms_cmk_regions import encrypt_decrypt
+from .examples_test_utils import get_cmk_arn
+from .examples_test_utils import static_plaintext
+
+
+pytestmark = [pytest.mark.examples]
+
+
+def test_one_kms_cmk():
```
ttjsu-aws: The name of this function doesn't accurately describe the functionality below. What are we trying to test here?

examples/test/test_i_multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,33 @@
...
+import botocore.session
+import pytest
+
+from ..src.multiple_kms_cmk_regions import multiple_kms_cmk_regions
+
+# from .examples_test_utils import get_cmk_arn
+from .examples_test_utils import static_plaintext
+
+
+pytestmark = [pytest.mark.examples]
```
karlw00t: Should this be used as a decorator?

mattsb42-aws: We could, but I'd prefer to keep it like this. `pytestmark` takes an iterable of markers and applies all of them to every test in the module.
In this particular case, there is only the one test, but I'd prefer to stick with this mechanism for consistency.

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    """Encrypts and then decrypts a string under multiple KMS customer master keys (CMKs) in multiple regions.
+
+    :param str key_arn1: Amazon Resource Name (ARN) of the KMS CMK
+    :param str key_arn2: Amazon Resource Name (ARN) of another KMS CMK
+    :param bytes source_plaintext: Data to encrypt
+    :param botocore_session: existing botocore session instance
+    :type botocore_session: botocore.session.Session
+    """
+    # Check that these keys are in different regions
+    assert not key_arn1[12:21] == key_arn2[12:21]
```
mattsb42-aws: This is brittle. Partition names are not always the same length and region names are not always the same length.

Arns are, however, always colon-delimited and ordered the same way, so you can split and pull out the region.

ex:
```python
region = key_arn1.split(":")[3]
```

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+Example showing basic encryption and decryption of a value already in memory
+using multiple KMS CMKs in multiple regions.
+"""
+import aws_encryption_sdk
+
+
+def encrypt(kms_key_provider, source_plaintext):
+    """Encrypts source_plaintext with the key(s) in kms_key_provider"""
+    return aws_encryption_sdk.encrypt(source=source_plaintext, key_provider=kms_key_provider)
```
mattsb42-aws: IMO, separating out the encrypt and decrypt logic at this level doesn't really make sense; you're basically just replacing a call to `aws_encryption_sdk.encrypt` with a call to `encrypt`.

@ttjsu-aws could you elaborate on where you meant to ask that the separation be drawn?

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    # Encrypt the plaintext using the AWS Encryption SDK. It returns the encrypted message and the header
+    ciphertext, encrypted_message_header = encrypt(kms_key_provider, source_plaintext)
+
+    # Check that both key ARNs are in the message headers
+    assert len(encrypted_message_header.encrypted_data_keys) == 2
+
+    # Decrypt the encrypted message using the AWS Encryption SDK. It returns the decrypted message and the header
+    # Either of our keys can be used to decrypt the message
+    plaintext1, decrypted_message_header1 = decrypt(
+        aws_encryption_sdk.KMSMasterKeyProvider(**dict(key_ids=[key_arn1])), ciphertext
```
mattsb42-aws: `KMSMasterKeyProvider` will decrypt anything you have permissions to, in any region. This construction does not guarantee that you are using the requested key. Instead, for this decrypt path, use `KMSMasterKey`. That will make sure that it only works with that single CMK.

Unrelated, this syntax is redundant: `call(**dict(key=value))`

The `**` operator unpacks a dictionary into named parameters. `dict()` builds and returns a dictionary using the provided named parameters.
Doing these in sequence in the same place is a no-op and just complicates the code.

ie, these are equivalent:
```python
aws_encryption_sdk.KMSMasterKeyProvider(**dict(key_ids=[key_arn1]))
aws_encryption_sdk.KMSMasterKeyProvider(key_ids=[key_arn1])
```

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    plaintext2, decrypted_message_header2 = decrypt(
+        aws_encryption_sdk.KMSMasterKeyProvider(**dict(key_ids=[key_arn2])), ciphertext
+    )
+
+    # Check that the original message and the decrypted message are the same
+    assert source_plaintext == plaintext1
```
mattsb42-aws: The output of the `decrypt` function will always be bytes.
`source_plaintext` might not be bytes.

Before this check, let's add a decoding if necessary:
```python
if not isinstance(source_plaintext, bytes):
    plaintext1 = plaintext1.decode("utf-8")
    plaintext2 = plaintext2.decode("utf-8")
```

examples/src/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    """Encrypts source_plaintext with the key(s) in kms_key_provider"""
+    return aws_encryption_sdk.encrypt(source=source_plaintext, key_provider=kms_key_provider)
+
+
+def decrypt(kms_key_provider, ciphertext):
+    """Decrypts ciphertext with the key(s) in kms_key_provider"""
+    return aws_encryption_sdk.decrypt(source=ciphertext, key_provider=kms_key_provider)
+
+
+def multiple_kms_cmk_regions(key_arn1, key_arn2, source_plaintext, botocore_session=None):
```
mattsb42-aws: [NP/Naming]
For consistency, please separate concepts using underscores (ditto for variable naming elsewhere):
```suggestion
def multiple_kms_cmk_regions(key_arn_1, key_arn_2, source_plaintext, botocore_session=None):
```

examples/test/test_i_multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_multiple_kms_cmk_regions():
+    plaintext = static_plaintext
+    cmk_arn1 = "arn:aws:kms:us-west-2:658956600833:alias/EncryptDecrypt"
```
mattsb42-aws: Rather than hard-coding these, add another helper in `integration_test_utils` and trickle that up through.
Also, when you do that, update the integration test readme.

Please use this environment variable: `AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID_2`.

I'll update our CI configurations to have this set to the correct value.
Related, the value that is stored here *must* be a CMK Arn, not an alias Arn (I'll make sure it is in CI).
We might relax that requirement once we switch over to keyrings, but for now various things require it.

examples/src/multiple_kms_cmk_regions.py
caitlin-tibbetts: Does this make the assumption that `source_plaintext` is either `bytes` or `str`?

mattsb42-aws:
:/ I think I may have broken this PR by changing it to point at `keyring`. I was trying to get this merged before I started working through the the #156 work for the keyring release, but that doesn't look like it'll work now.

I'll see if I can merge this into my examples branch and still retain the authorship.

mattsb42-aws:
That does appear to have worked. :)

I'll fix up the things that were blocking this merging and incorporate it into my examples revamp. I'll close this PR once I cut that one.

mattsb42-aws:
This is now merged in #219.

Thanks @caitlin-tibbetts and sorry this took so long. :)

Issue #31, opened 2017-12-02
some tests break when default region is set in .aws/config


mattsb42-aws:
## Problem
Some of our unit tests fail when there is a region set for the default profile in `.aws/config`.

PR #179, opened 2019-07-27
Fixed KMS master key provider tests when default AWS region is configured

ragona:
This fixes issue #31, in which users who have configured a default AWS region (either via ~/.aws/config or via environment variable) will experience failing tests for `KMSMasterKeyProvider`. I tried to keep the tests mostly unchanged, but in some cases I elected to simply skip them since the test didn't provide much value if there was already a default value configured. 

### DIFF
--- a/test/unit/test_providers_kms_master_key_provider.py
+++ b/test/unit/test_providers_kms_master_key_provider.py
@@ -23,6 +23,19 @@ from aws_encryption_sdk.key_providers.kms import KMSMasterKey, KMSMasterKeyProvi
 pytestmark = [pytest.mark.unit, pytest.mark.local]
 
 
+@pytest.fixture(autouse=True, params=[True, False], ids=["default region", "no default region"])
+def patch_default_region(request, monkeypatch):
+    """Run all tests in this module both with a default region set and no default region set.
+
+    This ensures that we do not regress on default region handling.
+    https://github.com/aws/aws-encryption-sdk-python/issues/31
+    """
+    if request.param:
+        monkeypatch.setenv("AWS_DEFAULT_REGION", "us-west-2")
+    else:
+        monkeypatch.delenv("AWS_DEFAULT_REGION", raising=False)
+
+
 def test_init_with_regionless_key_ids_and_region_names():
     key_ids = ("alias/key_1",)
     region_names = ("test-region-1",)
--- a/test/unit/test_providers_kms_master_key_provider.py
+++ b/test/unit/test_providers_kms_master_key_provider.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 """Unit test suite from aws_encryption_sdk.key_providers.kms.KMSMasterKeyProvider"""
 import botocore.client
+import botocore.session
 import pytest
 from mock import ANY, MagicMock, call, patch, sentinel
 
@@ -32,6 +33,7 @@ def test_init_with_regionless_key_ids_and_region_names():
 class TestKMSMasterKeyProvider(object):
     @pytest.fixture(autouse=True)
     def apply_fixtures(self):
+        self.botocore_no_region_session = botocore.session.Session(session_vars={"region": (None, None, None, None)})
         self.mock_botocore_session_patcher = patch("aws_encryption_sdk.key_providers.kms.botocore.session.Session")
         self.mock_botocore_session = self.mock_botocore_session_patcher.start()
         self.mock_boto3_session_patcher = patch("aws_encryption_sdk.key_providers.kms.boto3.session.Session")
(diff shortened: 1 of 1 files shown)


test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -124,9 +129,11 @@ def test_client_no_region_name_with_default(self, mock_add_client):
         test._regional_clients[sentinel.default_region] = sentinel.default_client
         client = test._client("")
         assert client is sentinel.default_client
-        mock_add_client.assert_called_once_with(sentinel.default_region)
+        mock_add_client.assert_called_with(sentinel.default_region)
 
     def test_client_no_region_name_without_default(self):
+        if default_region_is_set():
```
mattsb42-aws: The intent of this test is to verify what happens if a client is created that cannot determine the default region, and then we attempt to obtain a client using a key ID that is not a complete Arn.

Some better options for fixing this test would be one of:
* use the unknown-region botocore session mentioned earlier
* manually set the instance's default region to `None` before calling `test._client`
* two parts:
    1. change this test to instead mock out `_region_from_key_id` and then simply verify that it is called with the expected values
    1. add stand-alone tests for the `_region_from_key_id` function

test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -32,6 +33,7 @@ def test_init_with_regionless_key_ids_and_region_names():
 class TestKMSMasterKeyProvider(object):
     @pytest.fixture(autouse=True)
     def apply_fixtures(self):
+        self.botocore_no_region_session = botocore.session.Session(session_vars={"region": (None, None, None, None)})
```
ragona: This syntax is weird -- the reason it wants this tuple of values for region is a little obscure, but it's due to how botocore's `session.py` unpacks the config values in `_update_config_store_from_session_vars`. I tried a bunch of variants of this and this is what worked, but I'm not sure what the right convention is to comment this bit of weirdness. Super open to feedback here.

test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -70,17 +70,20 @@ def test_init_with_region_names(self, mock_add_clients):
     @patch("aws_encryption_sdk.key_providers.kms.KMSMasterKeyProvider.add_regional_client")
     def test_init_with_default_region_found(self, mock_add_regional_client):
         test = KMSMasterKeyProvider()
-        assert test.default_region is None
+        if not default_region_is_set():
+            assert test.default_region is None
```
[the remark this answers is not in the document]
ragona: Fixed this by creating a session with no default region.

test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -141,3 +148,10 @@ def test_new_master_key(self, mock_client):
         key = test._new_master_key(key_info)
         check_key = KMSMasterKey(key_id=key_info, client=self.mock_boto3_client_instance)
         assert key != check_key
+
+
+def default_region_is_set():
+    """Returns whether a user has their default AWS region set. This is useful for
+       testing logic around setting the region; there are some assertions which assume
+       the region starts out as `None`."""
+    return KMSMasterKeyProvider().default_region is not None
```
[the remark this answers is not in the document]
ragona: This method turned out to not be needed, so I removed it in the latest commit.

test/unit/test_providers_kms_master_key_provider.py
```diff
@@ -70,17 +70,20 @@ def test_init_with_region_names(self, mock_add_clients):
...
         ) as mock_get_config:
             test._process_config()
             mock_get_config.assert_called_once_with("region")
             assert test.default_region is sentinel.default_region
-            mock_add_regional_client.assert_called_once_with(sentinel.default_region)
+            mock_add_regional_client.assert_called_with(sentinel.default_region)
 
     @patch("aws_encryption_sdk.key_providers.kms.KMSMasterKeyProvider.add_regional_client")
     def test_init_with_default_region_not_found(self, mock_add_regional_client):
+        if default_region_is_set():
```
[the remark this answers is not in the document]
ragona: Fixed this by creating a session with no default region.

test/unit/test_providers_kms_master_key_provider.py
ragona: Fixed this by creating a session with no default region.

ragona:
Thanks for the very helpful comments, @mattsb42-aws! I think I've responded to all of your comments with a much-improved way to fix this that doesn't involve skipping any tests. I'm not sure if I followed the general best practices and conventions for creating that fixture though, so definitely looking for feedback on general style.

test/unit/test_providers_kms_master_key_provider.py
mattsb42-aws: I don't love it either, but as we discussed offline, I don't know that there's a better way to do it. This is fine.

Issue #135, opened 2019-01-04
DefaultCryptoMaterialsManager should raise an error if signer key is present in encryption context


mattsb42-aws:
Because `DefaultCryptoMaterialsManager` adds a value to the encryption context if a signed algorithm suite is used, we need to throw an error if the requested encryption context already contains that key.

https://github.com/aws/aws-encryption-sdk-java/blob/eb8702e82759cca53eddc3b5f8423e74ec31aff1/src/main/java/com/amazonaws/encryptionsdk/DefaultCryptoMaterialsManager.java#L55-L58

18AdrianoH:
Submitted PR.

https://github.com/aws/aws-encryption-sdk-python/pull/185

acioc:
Closed stale PR for this issue

PR #185, opened 2019-08-07
Issue135 (Cannot add key to Encryption Context if already there in default CMM)

18AdrianoH:
Deals with Issue #135 

Added a test to see whether it was possible to add another key with the same name or not.  The expected correct behavior is that an error will be thrown in this case (for more context read he issue).  Then fixed the issue and it works.  It ignores the value of the key, and only looks at the name.

Please only look at the most recent commit fixing issue135.  The others are left over from when I branched in my fork after fixing a Sphinx issue.

# Check any applicable:
No files moved, deleted, or added, but they were edited.

### DIFF
--- a/examples/src/one_kms_cmk_unsigned.py
+++ b/examples/src/one_kms_cmk_unsigned.py
@@ -1,0 +1,51 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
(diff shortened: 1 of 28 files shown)


18AdrianoH:
> What is the difference between this PR and #187?

There is some overlap because it split from master at the same time as the other issue (basically I fixed the other issue in master and then decided it would be cleaner to put it in its own branch but didn't hard reset before splitting and editing in this branch, so it carries over some commits.  Resetting now would delete the additional commits that deal with issue 135.  This is my mistake and I can handle it manually if necessary but the functionality should not be changed if you merge this like nothing happened.)

PR #187 fixes a sphinx linking issue for our docs.  This PR (does that and also) fixes an issue where keys are already present in an encryption context.

acioc:
Closing stale PR, please reopen as needed

Issue #55, opened 2018-06-19
Serialization of encryption context needs to check for single too-long key or value


david-koenig:
Serialization format of encryption context does not allow any individual key or value to be longer than 2^16-1 bytes. Does serialization code check for this limitation? Do we have test vectors that detect this issue?

mattsb42-aws:
Note for reference: this will implicitly throw an error when we try to serialize the value. We should instead throw a `SerializationError`.

```python
>>> import struct
>>> struct.pack('>H', 2**16 - 1)
b'\xff\xff'
>>> struct.pack('>H', 2**16)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
struct.error: 'H' format requires 0 <= number <= 65535
```

18AdrianoH:
I've fixed this issue and made a PR at https://github.com/aws/aws-encryption-sdk-python/pull/188

PR #188, opened 2019-08-15
Fixed issue #55

18AdrianoH:
Issue #55 

Serialization of encryption context will throw the correct type of error.  Also broke the serialization function into two using a helper function to improve readability and actually pass flake8 tests (which otherwise fail because of "too complex" which basically means it has too much if/else try/catch logic going on).

# Check any applicable:
No files moved or created, but they were edited.  This also contains a merge with Caitlin's PR so it is up to date.

### DIFF
--- a/src/aws_encryption_sdk/internal/formatting/encryption_context.py
+++ b/src/aws_encryption_sdk/internal/formatting/encryption_context.py
@@ -47,17 +47,13 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
     return struct.pack(fmt, message_id, aad_content_string.value, seq_num, length)
 
 
-def _key_or_value_too_large(encryption_context_list):
-    """Helper method to check whether a key or value are too long."""
-    longest_key = max(encryption_context_list, key=lambda x: len(x[0]))[0]
-    longest_value = max(encryption_context_list, key=lambda x: len(x[1]))[1]
-    return len(longest_key) > 2 ** 16 - 1 or len(longest_value) > 2 ** 16 - 1
-
-
 def _serialize_encryption_context_list(encryption_context_list, serialized_context):
     """Helper function to serialize the list of encryption context key-value pairs."""
     for key, value in sorted(encryption_context_list, key=lambda x: x[0]):
         try:
+            if len(key) > 2 ** 16 - 1 or len(value) > 2 ** 16 - 1:
+                raise SerializationError("Key or Value are too large.  Maximum length is 65535")
+
             serialized_context.extend(
                 struct.pack(
                     ">H{key_size}sH{value_size}s".format(key_size=len(key), value_size=len(value)),
@@ -67,19 +63,19 @@ def _serialize_encryption_context_list(encryption_context_list, serialized_conte
                     value,
                 )
             )
-        # we check to make sure that we return the right type of error message for an overly long key or value
         except struct.error as struct_error:
             message = str(struct_error)
-            # a key or value were too long and struct gave us the expected error
             if message == "'H' format requires 0 <= number <= 65535":
                 raise SerializationError("Key or Value are too large.  Maximum length is 65535")
-            # a key or value were too long and struct's error message is different (maybe an update)
-            if _key_or_value_too_large(encryption_context_list):
-                raise SerializationError("Key or Value are too large.  Maximum length is 65535")
-            # else we can just raise the struct error as it was (unknown)
-            raise SerializationError("Unknown Error with struct pack.  Could not serialize key or value")
+
+            # else unknown
+            raise SerializationError(
+                [SerializationError("Unknown Error with struct pack.  Could not serialize key or value"), struct_error]
+            )
+
         if len(serialized_context) > aws_encryption_sdk.internal.defaults.MAX_BYTE_ARRAY_SIZE:
             raise SerializationError("The serialized context is too large.")
+
     return bytes(serialized_context)
 
 
--- a/src/aws_encryption_sdk/internal/formatting/encryption_context.py
+++ b/src/aws_encryption_sdk/internal/formatting/encryption_context.py
@@ -82,15 +82,25 @@ def serialize_encryption_context(encryption_context):
             )
 
     for key, value in sorted(encryption_context_list, key=lambda x: x[0]):
-        serialized_context.extend(
-            struct.pack(
-                ">H{key_size}sH{value_size}s".format(key_size=len(key), value_size=len(value)),
-                len(key),
-                key,
-                len(value),
-                value,
+        try:
+            serialized_context.extend(
+                struct.pack(
+                    ">H{key_size}sH{value_size}s".format(key_size=len(key), value_size=len(value)),
+                    len(key),
+                    key,
+                    len(value),
+                    value,
+                )
             )
-        )
+        # we check to make sure that we return the right type of error message for an overly long key or value
+        except struct.error as struct_error:
+            message = str(struct_error)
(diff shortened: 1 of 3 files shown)


src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -47,6 +47,33 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
...
+                    len(key),
+                    key,
+                    len(value),
+                    value,
+                )
+            )
+        # we check to make sure that we return the right type of error message for an overly long key or value
+        except struct.error as struct_error:
+            message = str(struct_error)
+            if message == "'H' format requires 0 <= number <= 65535":
```
lizroth: Minor: This is a risky way to match. If in the future the `str()` output changes even slightly, this check will no longer pass. Given that we're trying to make trapping this a useful part of our UI/UX, I would either do a check instead on a substring, add a unit test that the `"'H' format requires 0 <= number <= 65535"` message still appears when we expect it to, or both.

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -47,6 +47,33 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
...
+                # the key or value were too large
+                raise SerializationError("Key or Value are too large.  Maximum length is 65535")
+            # else we can just raise the struct error as it was (unknown)
+            raise struct_error
```
lizroth: Is there some reason not to wrap this in a `SerializationError` for consistent behavior for our clients?

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -47,6 +47,33 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
     return struct.pack(fmt, message_id, aad_content_string.value, seq_num, length)
 
 
+# help function to simplify code
```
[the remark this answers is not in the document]
18AdrianoH: done

test/unit/test_encryption_context.py
```diff
@@ -190,3 +190,12 @@ def test_deserialize_encryption_context_empty(self):
             serialized_encryption_context=b""
         )
         assert test == {}
+
+    # these three tests (in one function) make sure that whether a key or value or both are too long
```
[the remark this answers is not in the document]
18AdrianoH: done.

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: added "unknown" serialization error if we do not match the key/value too large

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: did both

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -47,9 +47,15 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
...
-# this is necessary to pass flake8
+def _key_or_value_too_large(encryption_context_list):
+    """Helper method to check whether a key or value are too long."""
+    longest_key = max(encryption_context_list, key=lambda x: len(x[0]))[0]
+    longest_value = max(encryption_context_list, key=lambda x: len(x[1]))[1]
+    return len(longest_key) > 2 ** 16 - 1 or len(longest_value) > 2 ** 16 - 1
+
+
 def _serialize_encryption_context_list(encryption_context_list, serialized_context):
+    """Helper function to serialize the list of encryption context key-value pairs."""
```
lizroth: Sorry, I offset some lines a bit when I was making that comment and picked the wrong example.

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -47,9 +47,15 @@ def assemble_content_aad(message_id, aad_content_string, seq_num, length):
     return struct.pack(fmt, message_id, aad_content_string.value, seq_num, length)
 
 
-# help function to simplify code
-# this is necessary to pass flake8
+def _key_or_value_too_large(encryption_context_list):
+    """Helper method to check whether a key or value are too long."""
+    longest_key = max(encryption_context_list, key=lambda x: len(x[0]))[0]
```
lizroth: You are re-walking the list of keys and values to make this computation, but the list is already being walked at line 59. This is invoked at line 77 in an `except` block, which means you already are on a key or a value that is too large. Rather than re-walking the entire list, why not either 1) validate each item in line with the iteration on 59 or 2) check which of the current value(s) are too large in the `except`?

If you do want to walk the whole list, wouldn't it be better to do that right away to avoid wasting any work in the loop on 59?

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -64,11 +70,14 @@ def _serialize_encryption_context_list(encryption_context_list, serialized_conte
...
+            # a key or value were too long and struct gave us the expected error
             if message == "'H' format requires 0 <= number <= 65535":
-                # the key or value were too large
+                raise SerializationError("Key or Value are too large.  Maximum length is 65535")
+            # a key or value were too long and struct's error message is different (maybe an update)
+            if _key_or_value_too_large(encryption_context_list):
                 raise SerializationError("Key or Value are too large.  Maximum length is 65535")
             # else we can just raise the struct error as it was (unknown)
-            raise struct_error
+            raise SerializationError("Unknown Error with struct pack.  Could not serialize key or value")
```
lizroth: Why not pass through the `struct_error` as well? Why squash that info?

src/aws_encryption_sdk/internal/formatting/encryption_context.py
```diff
@@ -64,11 +70,14 @@ def _serialize_encryption_context_list(encryption_context_list, serialized_conte
         # we check to make sure that we return the right type of error message for an overly long key or value
         except struct.error as struct_error:
             message = str(struct_error)
+            # a key or value were too long and struct gave us the expected error
```
lizroth: I don't think this comment clarifies anything -- what do you think?

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: yup flake8 enforced this

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: Added the check before packing.  Alternatively I could put this in the first run-through since there are two loops to fail faster.  The other part of the code deals with other exceptions, however, and I think it should be ok to leave it in the second loop for clarity.

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: The error now contains two errors.  The message as above in a SerializationError and the struct error.

src/aws_encryption_sdk/internal/formatting/encryption_context.py
18AdrianoH: deleted and other unhelpful commenting

acioc:
Closing stale PR, please reopen as needed

Issue #211, opened 2020-02-20
[keyring] Enable use of keyrings


mattsb42-aws:
We have some keyrings defined, but we can't actually use them yet.

Following on from #209, I think that trying to build an abstraction layer that lets master key providers (MKPs) pretend to be keyrings is not viable. There are just too many subtle differences in the APIs, including:
- MKPs accept the plaintext length and plaintext body as input on encrypt. Keyrings do not. We've been trying to discourage this usage for a while, but we can't just cut it off without potentially breaking custom MKPs that might use these values.
- MKPs implicitly assume that they are generating the data key. This could work out ok if you never combine the MKP keyring wrapper with any other keyrings, but if you ever do then the potential issues expand significantly.

Instead, we're just going to expand the default CMM to accept either an MKP or a keyring, and contain the complexity there.

Testing this will require:

- [x] `DefaultCryptoMaterialsManager` MUST accept either a MKP or a keyring and require exactly one
- [x] `DefaultCryptoMaterialsManager` MUST fail if the keyring could not complete the cryptographic materials
- [x] `DefaultCryptoMaterialsManager` MUST fail if the keyring changes the algorithm, encryption context, or signing/verification key from the requested materials
- [x] `CachingCryptoMaterialsManager` MUST accept either a MKP or a keyring and require exactly one
- [x] `StreamEncryptor` MUST accept either a MKP or a keyring and require exactly one
- [x] `StreamDecryptor` MUST accept either a MKP or a keyring and require exactly one
- [x] All functional and integration tests must test with both MKPs and keyrings
- [x] raw MKPs and keyrings MUST inter-operate end-to-end
- [x] AWS KMS MKPs and keyrings MUST interoperate end-to-end

mattsb42-aws:
Testing this will require:

- [ ] `DefaultCryptoMaterialsManager` MUST accept either a MKP or a keyring
- [ ] `DefaultCryptoMaterialsManager` MUST require exactly one of same
- [ ] All functional and integration tests must test with both MKPs and keyrings

PR #216, opened 2020-02-25
feat: enable use of keyrings

mattsb42-aws:
*Issue #, if available:* #211 

*Description of changes:*

This adds support for using keyrings to `DefaultCryptographicMaterialsManager` as well as the appropriate passthroughs via `CachingCryptographicMaterialsManager` and the encrypt/decrypt entry points.

# Check any applicable:

### DIFF
--- a/src/aws_encryption_sdk/keyrings/base.py
+++ b/src/aws_encryption_sdk/keyrings/base.py
@@ -23,6 +23,8 @@ except ImportError:  # pragma: no cover
     # We only actually need these imports when running the mypy checks
     pass
 
+__all__ = ("Keyring",)
+
 
 class Keyring(object):
     """Parent interface for Keyring classes.
--- a/src/aws_encryption_sdk/keyrings/base.py
+++ b/src/aws_encryption_sdk/keyrings/base.py
@@ -26,10 +26,9 @@ class Keyring(object):
         # type: (EncryptionMaterials) -> EncryptionMaterials
         """Generate a data key if not present and encrypt it using any available wrapping key.
 
-        :param encryption_materials: Encryption materials for the keyring to modify.
-        :type encryption_materials: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :param EncryptionMaterials encryption_materials: Encryption materials for keyring to modify.
         :returns: Optionally modified encryption materials.
-        :rtype: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :rtype: EncryptionMaterials
         :raises NotImplementedError: if method is not implemented
         """
         raise NotImplementedError("Keyring does not implement on_encrypt function")
@@ -38,12 +37,10 @@ class Keyring(object):
         # type: (DecryptionMaterials, Iterable[EncryptedDataKey]) -> DecryptionMaterials
         """Attempt to decrypt the encrypted data keys.
 
-        :param decryption_materials: Decryption materials for the keyring to modify.
-        :type decryption_materials: aws_encryption_sdk.materials_managers.DecryptionMaterials
-        :param encrypted_data_keys: List of encrypted data keys.
-        :type: Iterable of :class:`aws_encryption_sdk.structures.EncryptedDataKey`
+        :param DecryptionMaterials decryption_materials: Decryption materials for keyring to modify.
+        :param List[EncryptedDataKey] encrypted_data_keys: List of encrypted data keys.
         :returns: Optionally modified decryption materials.
-        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :rtype: DecryptionMaterials
         :raises NotImplementedError: if method is not implemented
         """
         raise NotImplementedError("Keyring does not implement on_decrypt function")
--- a/src/aws_encryption_sdk/keyrings/base.py
+++ b/src/aws_encryption_sdk/keyrings/base.py
@@ -1,15 +1,5 @@
-# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License"). You
-# may not use this file except in compliance with the License. A copy of
-# the License is located at
-#
-# http://aws.amazon.com/apache2.0/
-#
-# or in the "license" file accompanying this file. This file is
-# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
-# ANY KIND, either express or implied. See the License for the specific
-# language governing permissions and limitations under the License.
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
 """Base class interface for Keyrings."""
 from aws_encryption_sdk.materials_managers import (  # only used for mypy; pylint: disable=unused-import
     DecryptionMaterials,
--- a/src/aws_encryption_sdk/keyrings/raw.py
+++ b/src/aws_encryption_sdk/keyrings/raw.py
@@ -9,9 +9,7 @@ import six
 from attr.validators import instance_of, optional
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
-from cryptography.hazmat.primitives.asymmetric import rsa
-from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey  # noqa pylint: disable=unused-import
-from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey  # noqa pylint: disable=unused-import
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey, RSAPublicKey
 
 from aws_encryption_sdk.exceptions import GenerateKeyError
 from aws_encryption_sdk.identifiers import EncryptionKeyType, KeyringTraceFlag, WrappingAlgorithm
@@ -223,21 +221,23 @@ class RawRSAKeyring(Keyring):
 
     :param str key_namespace: String defining the keyring ID
     :param bytes key_name: Key ID
-    :param RSAPrivateKey private_wrapping_key: Private encryption key with which to wrap plaintext data key (optional)
-    :param RSAPublicKey public_wrapping_key: Public encryption key with which to wrap plaintext data key (optional)
+    :param cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey private_wrapping_key:
+        Private encryption key with which to wrap plaintext data key (optional)
+    :param cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey public_wrapping_key:
+        Public encryption key with which to wrap plaintext data key (optional)
     :param WrappingAlgorithm wrapping_algorithm: Wrapping Algorithm with which to wrap plaintext data key
     :param MasterKeyInfo key_provider: Complete information about the key in the keyring
 
     .. note::
 
-       At least one of public wrapping key or private wrapping key must be provided.
+        At least one of public wrapping key or private wrapping key must be provided.
     """
 
     key_namespace = attr.ib(validator=instance_of(six.string_types))
     key_name = attr.ib(validator=instance_of(six.binary_type))
     _wrapping_algorithm = attr.ib(repr=False, validator=instance_of(WrappingAlgorithm))
-    _private_wrapping_key = attr.ib(default=None, repr=False, validator=optional(instance_of(rsa.RSAPrivateKey)))
-    _public_wrapping_key = attr.ib(default=None, repr=False, validator=optional(instance_of(rsa.RSAPublicKey)))
+    _private_wrapping_key = attr.ib(default=None, repr=False, validator=optional(instance_of(RSAPrivateKey)))
+    _public_wrapping_key = attr.ib(default=None, repr=False, validator=optional(instance_of(RSAPublicKey)))
 
     def __attrs_post_init__(self):
         # type: () -> None
--- a/src/aws_encryption_sdk/keyrings/raw.py
+++ b/src/aws_encryption_sdk/keyrings/raw.py
@@ -10,7 +10,8 @@ from attr.validators import instance_of, optional
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
-from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey, RSAPublicKey
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey  # noqa pylint: disable=unused-import
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey  # noqa pylint: disable=unused-import
 
 from aws_encryption_sdk.exceptions import GenerateKeyError
 from aws_encryption_sdk.identifiers import EncryptionKeyType, KeyringTraceFlag, WrappingAlgorithm
@@ -19,14 +20,14 @@ from aws_encryption_sdk.internal.formatting.deserialize import deserialize_wrapp
 from aws_encryption_sdk.internal.formatting.serialize import serialize_raw_master_key_prefix, serialize_wrapped_key
 from aws_encryption_sdk.key_providers.raw import RawMasterKey
 from aws_encryption_sdk.keyrings.base import Keyring
-from aws_encryption_sdk.structures import (  # pylint: disable=unused-import
+from aws_encryption_sdk.structures import (
     EncryptedDataKey,
     KeyringTrace,
     MasterKeyInfo,
     RawDataKey,
 )
 
-from aws_encryption_sdk.materials_managers import (  # only used for mypy; pylint: disable=unused-import
+from aws_encryption_sdk.materials_managers import (
     DecryptionMaterials,
     EncryptionMaterials,
 )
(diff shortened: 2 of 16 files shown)


mattsb42-aws:
reminder for the record: Appveyor failing on 3.4 is a known issue and not caused by this PR

src/aws_encryption_sdk/materials_managers/caching.py
```diff
@@ -59,17 +56,17 @@ class CachingCryptoMaterialsManager(CryptoMaterialsManager):
...
-
-    :param cache: Crypto cache to use with material manager
-    :type cache: aws_encryption_sdk.caches.base.CryptoMaterialsCache
-    :param backing_materials_manager: Crypto material manager to back this caching material manager
-        (either `backing_materials_manager` or `master_key_provider` required)
-    :type backing_materials_manager: aws_encryption_sdk.materials_managers.base.CryptoMaterialsManager
-    :param master_key_provider: Master key provider to use (either `backing_materials_manager` or
-        `master_key_provider` required)
-    :type master_key_provider: aws_encryption_sdk.key_providers.base.MasterKeyProvider
+        Either ``backing_materials_manager``, ``keyring``, or ``master_key_provider`` must be provided.
```
seebees: if you combine multiple should that be an error?

src/aws_encryption_sdk/materials_managers/caching.py
```diff
@@ -111,10 +101,21 @@ def __attrs_post_init__(self):
         if self.max_age <= 0.0:
             raise ValueError("max_age cannot be less than or equal to 0")
 
+        options_provided = [
+            option is not None for option in (self.backing_materials_manager, self.keyring, self.master_key_provider)
+        ]
+        provided_count = len([is_set for is_set in options_provided if is_set])
+
+        if provided_count != 1:
+            raise TypeError("Exactly one of 'materials_manager', 'keyring', or 'key_provider' must be provided")
```
seebees: Above you say that you prefer `backing_materials_manager`.
But here you error
and call it `materials_manager`

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -101,6 +107,64 @@ def get_encryption_materials(self, request):
...
+        :returns: encryption materials
+        :rtype: aws_encryption_sdk.materials_managers.EncryptionMaterials
+        :raises InvalidCryptographicMaterialsError: if keyring cannot complete encryption materials
+        :raises InvalidCryptographicMaterialsError:
+           if encryption materials received from keyring do not match request
+        :raises MasterKeyProviderError: if no master keys are available from the underlying master key provider
+        :raises MasterKeyProviderError: if the primary master key provided by the underlying master key provider
+            is not included in the full set of master keys provided by that provider
+        """
+        if self.master_key_provider is not None:
```
seebees: If we are being paranoid.  Do we need to check that there is no keyring here?
I don't know how mutable the python object is...(?)

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -101,6 +107,64 @@ def get_encryption_materials(self, request):
...
+        :raises InvalidCryptographicMaterialsError: if keyring cannot complete encryption materials
+        :raises InvalidCryptographicMaterialsError:
+           if encryption materials received from keyring do not match request
+        """
+        algorithm = request.algorithm if request.algorithm is not None else self.algorithm
+        encryption_context = request.encryption_context.copy()
+
+        signing_key = self._generate_signing_key_and_update_encryption_context(algorithm, encryption_context)
+
+        expected_encryption_context = encryption_context.copy()
```
seebees: Why not do the copy the same way you do below?
i.e. copy on pass?

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -101,6 +107,64 @@ def get_encryption_materials(self, request):
...
+        encryption_materials = EncryptionMaterials(
+            algorithm=algorithm, encryption_context=encryption_context, signing_key=signing_key,
+        )
+
+        final_materials = self.keyring.on_encrypt(encryption_materials=encryption_materials)
+
+        if not final_materials.is_complete:
+            raise InvalidCryptographicMaterialsError("Encryption materials are incomplete!")
+
+        materials_are_valid = (
```
seebees: Who validates that you have a plaintext data key?
And that it matches the suite that you have?
Is that in the materials?

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -101,6 +107,64 @@ def get_encryption_materials(self, request):
...
+            final_materials.algorithm is algorithm,
+            final_materials.encryption_context == expected_encryption_context,
+            final_materials.signing_key is signing_key,
```
seebees: You copy the EC, but not the signing_key?
Is this how python deals with objects vs "strings"?

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -143,3 +207,55 @@ def decrypt_materials(self, request):
...
+        )
+
+        final_materials = self.keyring.on_decrypt(
+            decryption_materials=decryption_materials, encrypted_data_keys=request.encrypted_data_keys
+        )
+
+        if not final_materials.is_complete:
+            raise InvalidCryptographicMaterialsError("Decryption materials are incomplete!")
+
+        materials_are_valid = (
```
seebees: Need to have a plaintext data key...

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -143,3 +207,55 @@ def decrypt_materials(self, request):
...
+
+        :param request: decrypt materials request
+        :type request: aws_encryption_sdk.materials_managers.DecryptionMaterialsRequest
+        :returns: decryption materials
+        :rtype: aws_encryption_sdk.materials_managers.DecryptionMaterials
+        :raises InvalidCryptographicMaterialsError: if keyring cannot complete decryption materials
+        :raises InvalidCryptographicMaterialsError:
+           if decryption materials received from keyring do not match request
+        """
+        verification_key = self._load_verification_key_from_encryption_context(
```
seebees: Does this throw if the suite wants a key, but the EC does not have one?

src/aws_encryption_sdk/streaming_client.py
```diff
@@ -83,28 +80,27 @@ class _ClientConfig(object):
...
-    materials_manager = attr.ib(
-        hash=True, default=None, validator=attr.validators.optional(attr.validators.instance_of(CryptoMaterialsManager))
-    )
-    key_provider = attr.ib(
-        hash=True, default=None, validator=attr.validators.optional(attr.validators.instance_of(MasterKeyProvider))
-    )
-    source_length = attr.ib(
-        hash=True, default=None, validator=attr.validators.optional(attr.validators.instance_of(six.integer_types))
-    )
+    materials_manager = attr.ib(hash=True, default=None, validator=optional(instance_of(CryptoMaterialsManager)))
```
seebees: Why is this changing?  It looks like it is just formating?

src/aws_encryption_sdk/streaming_client.py
```diff
@@ -83,28 +80,27 @@ class _ClientConfig(object):
...
+        provided_count = len([is_set for is_set in options_provided if is_set])
+
+        if provided_count != 1:
+            raise TypeError("Exactly one of 'materials_manager', 'keyring', or 'key_provider' must be provided")
 
-        if both_cmm_and_mkp_defined or neither_cmm_nor_mkp_defined:
-            raise TypeError("Exactly one of materials_manager or key_provider must be provided")
         if self.materials_manager is None:
-            self.materials_manager = DefaultCryptoMaterialsManager(master_key_provider=self.key_provider)
+            if self.key_provider is not None:
```
seebees: There was an error, but now there is not.
I'm not sure I like the nested if,
What about checking `A & !B` and then `!A & B`, else `raise`?

src/aws_encryption_sdk/keyrings/base.py
```diff
@@ -1,15 +1,5 @@
...
-# may not use this file except in compliance with the License. A copy of
-# the License is located at
-#
-# http://aws.amazon.com/apache2.0/
-#
-# or in the "license" file accompanying this file. This file is
-# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
-# ANY KIND, either express or implied. See the License for the specific
-# language governing permissions and limitations under the License.
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
mattsb42-aws: I changed the header for any files that I touched. I will be doing a big PR later that is just these header changes, but for this PR, rather than updating the year range I decided it would be simpler to just do the move now for those files.

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: This check is determining which path we need to take. The logic in `__attrs_post_init__` makes one of these always non-`None`.

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: Dictionaries are mutable; I try to make sure that we don't feed any side-effects upstream or cause any side-effects downstream.

The copy on line 122 is taking a copy of the requested encryption context so that we can modify it without changing the value in the request.

```python
encryption_context = request.encryption_context.copy()
```

Then, on line 124, we modify it. Long-term I would like to refactor that method so that it returns the new encryption context rather than mutating the input, but mutation is how it has always been so I'm leaving that as-is for now to avoid feature-creep.

```python
signing_key = self._generate_signing_key_and_update_encryption_context(algorithm, encryption_context)
```

On line 126, we take a copy of the modified encryption context for later comparison.

```python
expected_encryption_context = encryption_context.copy()
```

Finally, on line 139, we check the encryption context in the final encryption materials against the copy we made on line 126, before we passed it to the keyring, to make sure that the keyring did not modify it.

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: The materials do. That's what we're checking on line 134 where we ask the materials if they're complete.

https://github.com/aws/aws-encryption-sdk-python/blob/952bce44db630754de6a0dad451f72665f83b1c7/src/aws_encryption_sdk/materials_managers/__init__.py#L248-L264

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: Dictionaries are mutable; `bytes` are not.

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: `is_complete`

https://github.com/aws/aws-encryption-sdk-python/blob/952bce44db630754de6a0dad451f72665f83b1c7/src/aws_encryption_sdk/materials_managers/__init__.py#L404-L420

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: Yes.

https://github.com/aws/aws-encryption-sdk-python/blob/952bce44db630754de6a0dad451f72665f83b1c7/src/aws_encryption_sdk/materials_managers/default.py#L116-L117

src/aws_encryption_sdk/streaming_client.py
mattsb42-aws: I changed the validator references from their fully-qualified names to importing the validators directly to reduce boilerplate.

src/aws_encryption_sdk/streaming_client.py
mattsb42-aws: Now that we have three possible options, it's more complicated. That's what I'm doing in 93-97, where I throw an error if you supply anything other than exactly one of the possible options.

seebees:
My questions have been resolved.  @ajw-aws  :)

src/aws_encryption_sdk/keyrings/raw.py
```diff
@@ -20,6 +10,8 @@
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey  # noqa pylint: disable=unused-import
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey  # noqa pylint: disable=unused-import
```
ajw-aws: Why are we importing these if it is not used?

src/aws_encryption_sdk/keyrings/raw.py
```diff
@@ -240,19 +219,18 @@ class RawRSAKeyring(Keyring):
...
-    :type key_provider: MasterKeyInfo
+    :param RSAPrivateKey private_wrapping_key: Private encryption key with which to wrap plaintext data key (optional)
+    :param RSAPublicKey public_wrapping_key: Public encryption key with which to wrap plaintext data key (optional)
+    :param WrappingAlgorithm wrapping_algorithm: Wrapping Algorithm with which to wrap plaintext data key
+    :param MasterKeyInfo key_provider: Complete information about the key in the keyring
 
     .. note::
-    At least one of public wrapping key or private wrapping key must be provided.
+
+       At least one of public wrapping key or private wrapping key must be provided.
```
ajw-aws: nit: three spaces should be four

src/aws_encryption_sdk/materials_managers/caching.py
```diff
@@ -111,10 +100,21 @@ def __attrs_post_init__(self):
         if self.max_age <= 0.0:
             raise ValueError("max_age cannot be less than or equal to 0")
 
+        options_provided = [
+            option is not None for option in (self.backing_materials_manager, self.keyring, self.master_key_provider)
+        ]
+        provided_count = len([is_set for is_set in options_provided if is_set])
+
+        if provided_count != 1:
+            raise TypeError("Exactly one of 'backing_materials_manager', 'keyring', or 'key_provider' must be provided")
```
ajw-aws: Is TypeError the best exception we can throw here? Is there a better custom exception class, or should we create one?

src/aws_encryption_sdk/materials_managers/default.py
```diff
@@ -143,3 +207,55 @@ def decrypt_materials(self, request):
...
+            verification_key=verification_key,
+        )
+
+        final_materials = self.keyring.on_decrypt(
+            decryption_materials=decryption_materials, encrypted_data_keys=request.encrypted_data_keys
+        )
+
+        materials_are_valid = (
+            final_materials.algorithm is request.algorithm,
+            final_materials.encryption_context == request.encryption_context,
```
ajw-aws: Is there a reason we have to use `==` here instead of `is` ?

test/unit/materials_managers/test_caching.py
```diff
@@ -88,20 +80,26 @@ def test_custom_partition_name(patch_uuid4):
...
 
-def test_mkp_to_default_cmm(mocker):
-    mocker.patch.object(aws_encryption_sdk.materials_managers.caching, "DefaultCryptoMaterialsManager")
-    mock_mkp = MagicMock(__class__=MasterKeyProvider)
+def test_mkp_to_default_cmm():
+    mkp = ephemeral_raw_aes_master_key()
+
     test = CachingCryptoMaterialsManager(
-        cache=MagicMock(__class__=CryptoMaterialsCache), max_age=10.0, master_key_provider=mock_mkp
+        cache=LocalCryptoMaterialsCache(capacity=10), max_age=10.0, master_key_provider=mkp
```
ajw-aws: Where does the magic number `10` come from? Should this instead be a declared constant elsewhere?

src/aws_encryption_sdk/keyrings/raw.py
mattsb42-aws: Unfortunately not; Python is 4-space indent; rst is 3-space indent.

src/aws_encryption_sdk/materials_managers/caching.py
mattsb42-aws: `TypeError` is consistent with what built-ins and `attrs` throw if you provide incorrect parameters or types.

```python
>>> def foo(b):
...     print("foo")
... 
>>> foo(3)
foo
>>> foo(b=3)
foo
>>> foo(f=3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() got an unexpected keyword argument 'f'
>>> foo(3, 4)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() takes 1 positional argument but 2 were given

```

src/aws_encryption_sdk/materials_managers/default.py
mattsb42-aws: Because we send a copy into the keyring rather than the original to avoid any potential side effects from misbehaving keyrings modifying it.

test/unit/materials_managers/test_caching.py
mattsb42-aws: It's not magic, it's just a required parameter. It wasn't there before because I was using mocks before and now I'm actually using the actual class. IMO making it a constant would give it undue gravity; it really doesn't matter what number it is.

src/aws_encryption_sdk/keyrings/raw.py
mattsb42-aws: I stand corrected; I checked this while tweaking some other aspects of the docs and 4-space works here as well. There are some places in rst that require 3-space, but it looks like this is not one of them.

Issue #212, opened 2020-02-20
[keyring] add AWS KMS keyring


mattsb42-aws:
Add an AWS KMS keyring as defined in the specification: https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md

Issue #146, opened 2019-04-05
add keyrings


mattsb42-aws:
Starting with the AWS Encryption SDK for C, we moved from master keys/providers to keyrings as a configuration/framework primitive. We need to add corresponding keyrings to this client as well.

In order to maintain backwards compatibility, we need to retain the existing MKP framework and support the use of MKPs within the new keyring framework. After discussing several options, we decided that the best route to accomplish this is with a MKP keyring.

This MKP keyring will take a MKP as input and provide the MK/MKP management logic currently in `DefaultCryptoMaterialsManager`, along with necessary keyring logic (ex: adding keyring traces), all wrapped up in the new keyring interface.

As a temporary measure, in order to enable end-to-end keyring tests without tackling the MKP keyring and the updates to `DefaultCryptoMaterialsManager`, we will need to make a `KeyringDefaultCryptoMaterialsManager` that handles the signing/verification keys but only understands keyrings.

All keyring development PRs should be made to the `keyring` branch.

- [x] Updated cryptographic materials to meet keyring needs [pr: #163]
- [x] Make updated encryption materials immutable [issue: #214]
- [x] `KeyringTraceFlag` does not need to be value/name-compatible with C [issue: #215]
- [x] Keyring interface [pr: #161]
- [x] AWS KMS keyring [issue: #212] [pr: #218]
- [x] Raw RSA keyring [pr: #165]
- [x] Raw AES keyring [pr: #165]
- [x] Multi keyring [pr: #166]
- [x] Enable keyring use. [issue: #211] [pr: #216]
- [x] Migration guide for users [issue: #213]

PR #218, opened 2020-03-05
feat: add AWS KMS keyring

mattsb42-aws:
*Issue #, if available:* #146 #212 

*Description of changes:*

This adds the AWS KMS keyring as defined in the spec[1].

Because the non-discovery KMS keyring behavior is identical to the multi-keyring behavior, rather than building this keyring from scratch, I instead reused the multikeyring internally to provide expected behavior for non-discovery modes.

[1] https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md

# Check any applicable:


src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    Set ``child_key_ids`` to specify additional CMKs that the keyring will use to encrypt the data key.
+
+    The keyring will attempt to use any CMKs
+    identified by CMK ARN in either ``generator_key_id`` or ``child_key_ids`` on decrypt.
+
+    You can identify CMKs by any `valid key ID`_ for the keyring to use on encrypt,
+    but for the keyring to attempt to use them on decrypt
+    you MUST specify the CMK ARN.
+
+    If you specify neither ``generator_key_id`` nor ``child_key_ids``
```
seebees: When talking about this in the ESDK-JS we went with "explicit is better"
so you must pass `discovery : true`.
I'm not sure of the best way to do this in Python,
or if this is crazy.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    _grant_tokens = attr.ib(
+        default=attr.Factory(tuple),
+        validator=(deep_iterable(member_validator=instance_of(six.string_types)), value_is_not_a_string),
+    )
+
+    def on_encrypt(self, encryption_materials):
+        # type: (EncryptionMaterials) -> EncryptionMaterials
+        trace_info = MasterKeyInfo(provider_id=_PROVIDER_ID, key_info=self._key_id)
+        try:
+            if encryption_materials.data_encryption_key is None:
```
seebees: This makes the keyring rely on the `MultiKeyring` to enforce the generator relationship.
I think I'm ok with this,
but I at least want to bring it up.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    from typing import Any, Dict, Iterable, Union  # noqa pylint: disable=unused-import
+except ImportError:  # pragma: no cover
+    # We only actually need these imports when running the mypy checks
+    pass
+
+__all__ = ("KmsKeyring",)
+
+_LOGGER = logging.getLogger(__name__)
+_PROVIDER_ID = "aws-kms"
+_GENERATE_FLAGS = {KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY}
```
seebees: Should this also have Encrypted and Signed?

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+
+        encryption_materials.add_encrypted_data_key(
+            encrypted_data_key=encrypted_key, keyring_trace=KeyringTrace(wrapping_key=trace_info, flags=_ENCRYPT_FLAGS)
+        )
+
+        return encryption_materials
+
+    def on_decrypt(self, decryption_materials, encrypted_data_keys):
+        # type: (DecryptionMaterials, Iterable[EncryptedDataKey]) -> DecryptionMaterials
+        for edk in encrypted_data_keys:
```
seebees: There is something about this that I feel sad about.
I feel like there should be a more efficient way to search the EDKs list,
and then _only_ process the matching ones...

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+        # type: (EncryptionMaterials) -> EncryptionMaterials
+        return encryption_materials
+
+    def on_decrypt(self, decryption_materials, encrypted_data_keys):
+        # type: (DecryptionMaterials, Iterable[EncryptedDataKey]) -> DecryptionMaterials
+        for edk in encrypted_data_keys:
+            if decryption_materials.data_encryption_key is not None:
+                return decryption_materials
+
+            if edk.key_provider.provider_id != _PROVIDER_ID:
```
seebees: as with the comment above,
maybe what I'm thinking is that a filter would be positive logic,
which would be easier to read?

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    # type: (ClientSupplier, str, EncryptedDataKey, Dict[str, str], Iterable[str]) -> RawDataKey
+    """Attempt to call ``kms:Decrypt`` and return the resulting plaintext data key.
+
+    Any errors encountered are passed up the chain without comment.
+
+    .. versionadded:: 1.5.0
+
+    """
+    region = _region_from_key_id(encrypted_data_key.key_provider.key_info.decode("utf-8"))
+    client = client_supplier.client(region)
```
seebees: What if the supplier does not return a client?

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def _do_aws_kms_encrypt(client_supplier, key_name, plaintext_data_key, encryption_context, grant_tokens):
+    # type: (ClientSupplier, str, RawDataKey, Dict[str, str], Iterable[str]) -> EncryptedDataKey
+    """Attempt to call ``kms:Encrypt`` and return the resulting encrypted data key.
+
+    Any errors encountered are passed up the chain without comment.
+    """
+    region = _region_from_key_id(key_name)
+    client = client_supplier.client(region)
```
seebees: What if the supplier does not return a client?

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
```diff
@@ -0,0 +1,376 @@
...
+    # type: (ClientSupplier, str, Dict[str, str], AlgorithmSuite, Iterable[str]) -> (RawDataKey, EncryptedDataKey)
+    """Attempt to call ``kms:GenerateDataKey`` and return the resulting plaintext and encrypted data keys.
+
+    Any errors encountered are passed up the chain without comment.
+
+    .. versionadded:: 1.5.0
+
+    """
+    region = _region_from_key_id(key_name)
+    client = client_supplier.client(region)
```
seebees: In this case, no client is an error.

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
```diff
@@ -0,0 +1,136 @@
...
+    from typing import Any, Dict, Union  # noqa pylint: disable=unused-import
+except ImportError:  # pragma: no cover
+    # We only actually need these imports when running the mypy checks
+    pass
+
+_LOGGER = logging.getLogger(__name__)
+__all__ = ("ClientSupplier", "DefaultClientSupplier", "AllowRegionsClientSupplier", "DenyRegionsClientSupplier")
+
+
+class ClientSupplier(object):
```
seebees: nit: The specification talks about a function not a class.
If we truly care, maybe we should update the spec,
at least with our not caring?

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    allowed_regions = attr.ib(
+        validator=(deep_iterable(member_validator=instance_of(six.string_types)), value_is_not_a_string)
+    )
+    _botocore_session = attr.ib(default=attr.Factory(BotocoreSession), validator=instance_of(BotocoreSession))
+
+    def __attrs_post_init__(self):
+        """Set up internal client supplier."""
+        self._supplier = DefaultClientSupplier(botocore_session=self._botocore_session)
```
seebees: Maybe nobody cares. But how do I compose _my_ client supplier with this Allow one?
Should the `_supplier` also be an option along with the `allowed_regions`?
(Or am I not reading this correctly?)

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
```diff
@@ -0,0 +1,136 @@
...
+        # type: (Union[None, str]) -> BaseClient
+        """Return a client for the requested region.
+
+        :rtype: BaseClient
+        """
+        raise NotImplementedError("'ClientSupplier' does not implement 'client'")
+
+
+@attr.s
+class DefaultClientSupplier(ClientSupplier):
```
seebees: How do I build a supplier that has different options, like timeout?
I think that people generally will want to use the cache,
but is that always the case?

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
mattsb42-aws: I've been thinking about this, and I think what I'll do here is make the inner supplier an optional parameter for the filtering suppliers and the cache an optional parameter for the default supplier. That way you can plumb in whatever configurations you want by assembling the pieces.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
seebees: For now I'm ok with a derived property.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
seebees: Nice change to positive.

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
seebees: It is super simple to make your own.
I agree that the default should be to cache for almost all cases.

src/aws_encryption_sdk/keyrings/aws_kms/client_suppliers.py
seebees: addressed elsewhere.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
seebees: This means that if this `_AwsKmsSingleCmkKeyring` is used in a MultiKeyring
*not* in generator position,
and this keyring is used as a generator for a parent MultiKeyring,
This combined keyring will **not** "do the right thing".

Will talk more.
But I want this to get documented.

mattsb42-aws: That is addressed by the multi-keyring. That's one reason why I chose to use the multi-keyring as an internal construct; aside from discovery mode, the requirements are identical.

https://github.com/aws/aws-encryption-sdk-python/blob/71b795dccb280b2b25b92c359d80c85623f2bd27/src/aws_encryption_sdk/keyrings/multi.py#L63-L77

seebees: Yup, thanks.

src/aws_encryption_sdk/keyrings/aws_kms/__init__.py
seebees: This is ok, for now.
This does not block this release,
and mirrors the Java implementation.

Issue #213, opened 2020-02-21
[keyring] keyring migration guide


mattsb42-aws:
I'd like for us to have a migration guide for how to move from master key providers to keyrings. I'm not sure exactly what form this should take. The main things I want to make sure it does are:

- Show how to change from MKP to keyring. This is especially important for raw MKPs.
- Explain why you should change. ie, what do keyrings give you that MKPs do not?

mattsb42-aws:
Thinking through this, I think that the best way to show *how* to migrate is by creating examples of common use-cases using both keyrings and master key providers.

mattsb42-aws:
Apart from the examples, we decided that would fit best in the developer guide.

PR #219, opened 2020-03-09
Examples refresh, take 1

mattsb42-aws:
*Issue #, if available:*
#156
#213 
resolves: #177 

*Description of changes:*

There are three main things going on here:
1. Forcing every example author to also write a test is cumbersome. To get away from this, I reworked how we test the examples so that as long as the example function has the right name and signature (outlined in the new examples readme) it will Just Work.
1. In preparation for all of the revised examples, I pulled all of the existing examples into a `legacy` module. I'll be pulling some of them out into their appropriate place later, pending the next item.
    * As part of this, I also pulled in the example that @caitlin-tibbetts added in #177 to avoid churn.
1. I added some initial new-format examples to show how to use the high-level APIs. I want to make sure we like this format (guiding comments, layout, flow, etc) before writing the couple dozen more examples that we need so that I don't need to copy-paste the same changes into dozens of files.

# Check any applicable:
- [x] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.


examples/src/legacy/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,64 @@
...
+    # Check that these keys are in different regions
+    assert not aws_kms_generator_cmk.split(":")[3] == child_cmk.split(":")[3]
+
+    kwargs = dict(key_ids=[aws_kms_generator_cmk, child_cmk])
+
+    if botocore_session is not None:
+        kwargs["botocore_session"] = botocore_session
+
+    # Create master key provider using the ARNs of the keys and the session (botocore_session)
+    kms_key_provider = KMSMasterKeyProvider(**kwargs)
```
acioc: If botocore_session is None, is the `KMSMasterKeyProvider` step still safe to perform? E.g. Is a new session created? Does this fail and crash? etc.

examples/src/legacy/multiple_kms_cmk_regions.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Encrypt the plaintext using the AWS Encryption SDK. It returns the encrypted message and the header
+    ciphertext, encrypted_message_header = aws_encryption_sdk.encrypt(
+        key_provider=kms_key_provider, source=source_plaintext
+    )
+
+    # Check that both key ARNs are in the message headers
+    assert len(encrypted_message_header.encrypted_data_keys) == 2
```
acioc: Do we need to verify these are actually ARNs?

test/integration/README.rst
```diff
@@ -5,7 +5,8 @@ aws-encryption-sdk Integration Tests
 In order to run these integration tests successfully, these things must be configured.
 
 #. Ensure that AWS credentials are available in one of the `automatically discoverable credential locations`_.
-#. Set environment variable ``AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID`` to valid
+#. Set environment variable ``AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID``
+   and ``AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID_2`` to valid
```
acioc: Do we have documentation on `AWS_KMS_KEY_ID` vs `AWS_KMS_KEY_ID_2`? Are there times when we would want to set one to a value and 2 to a different value?

Edit: `integration_test_utils.py` makes this clearer. Would be good to also note why we have both here as well, though.

tox.ini
```diff
@@ -140,6 +141,10 @@ commands =
 
 [testenv:flake8-examples]
 basepython = {[testenv:flake8]basepython}
+# This does not actually ignore errors,
+#  it just runs all commands regardless of whether any fail.
+#  If any fail, the final result is still a fail.
```
acioc: np: extra whitespace?

np: `the final result is still a fail` -> `the final result will still fail`

tox.ini
```diff
@@ -164,6 +169,10 @@ commands =
 
 [testenv:pylint-examples]
 basepython = {[testenv:pylint]basepython}
+# This does not actually ignore errors,
+#  it just runs all commands regardless of whether any fail.
+#  If any fail, the final result is still a fail.
```
acioc: np: extra whitespace?

np: `the final result is still a fail` -> `the final result will still fail`

examples/README.md
```diff
@@ -0,0 +1,93 @@
+# AWS Encryption SDK Examples
+
+Here you can find some examples that show you
+how to use the AWS Encryption SDK.
+We demonstrate how to use the high-level APIs
+as well as how to set up some common configuration patterns.
```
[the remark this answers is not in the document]
mattsb42-aws: https://rhodesmill.org/brandon/2012/one-sentence-per-line/

The TLDR is: "insert a newline wherever you have a cognitive break." The idea is to keep lines to a reasonable length while also reducing the likelihood of content changes requiring changes to unrelated lines.

examples/src/legacy/multiple_kms_cmk_regions.py
mattsb42-aws: In my next-step cleanup, I'll be removing that as an option in the example (and probably adding an example of how to provide a custom session instead). If you don't provide a session, the KMS MKP will automatically create one for you.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+1. The example MUST be a distinct module in the [`examples/src/`](./src) directory.
+1. The example MAY be nested arbitrarily deeply,
+    but every intermediate directory MUST contain a `__init__.py` file
+    so that CPython 2.7 will recognize it as a module.
+1. Every example MUST be CPython 2.7 compatible.
+1. Each example file MUST contain exactly one example.
+1. Each example file MUST contain a function called `run` that runs the example.
+1. If your `run` function needs any of the following inputs,
+    the parameters MUST have the following names:
+    * `aws_kms_cmk_arn` (`str`) : A single AWS KMS CMK ARN.
```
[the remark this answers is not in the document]
mattsb42-aws: Fair point. I had originally used that name because that's what the previous examples used, but there's no reason we need to be stuck with that.

examples/src/legacy/multiple_kms_cmk_regions.py
mattsb42-aws: IMO that doesn't really add much to the example, but I'm willing to be convinced otherwise.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+Keyrings are the most common way for you to configure that AWS Encryption SDK.
+These let you define how you want the AWS Encryption SDK to protect your data keys.
+You can find these examples in [`examples/src/keyring`](./src/keyring).
+
+### Cryptographic Materials Managers
+
+Keyrings define how you want to protect your data keys,
+but there is more going on here than just data keys.
+
+Cryptographic materials managers give you higher-level controls
```
juneb: higher-level or lower-level? I think of the CMM as lower-level.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+over how the AWS Encryption SDK protects your data.
+This can include things like
+enforcing certain algorithm suites or encryption context settings,
```
juneb:
```suggestion
enforcing the use of certain algorithm suites or encryption context settings,
```

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+reusing data keys across messages,
+or changing how you interact with keyrings.
+You can find these examples in
+[`examples/src/crypto_materials_managers`](./src/crypto_materials_manager).
+
+### Master Key Providers
+
+Before there were keyrings, there were master key providers.
+Master key providers were the original configuration structure
+that we defined for defining how you want to protect your data keys.
```
juneb:
```suggestion
defined for defining ???
```

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+Keyrings provide a simpler experience and often more powerful configuration options,
+but if you need to use master key providers,
+need help migrating from master key providers to keyrings,
+or simply want to see the difference between these configuration experiences,
+you can find these examples in [`examples/src/master_key_provider`](./src/master_key_provider).
```
juneb: In Java, master keys/master key providers are deprecated. Not so in Python?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Legacy
+
+These are any examples that were already defined
+before we started revamping our examples.
```
juneb:
```suggestion
before we started revamping our examples.
```
```suggestion
This section includes older examples, including examples of using master keys and master key providers in Java and Python.
```

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+We are keeping them around for anyone who needs them as reference material,
```
juneb:
```suggestion
You can use them as a reference,
```

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+but we recommend looking at the newer examples
+that should provide a clearer picture of how to use this library.
```
juneb:
```suggestion
but we recommend looking at the newer examples, which explain the preferred ways of using this library.
```

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+You can find these examples in [`examples/src/legacy`](./src/legacy).
+
+# Writing Examples
+
+If you want to write a new example, that's awesome!
```
juneb:
```suggestion
If you want to contribute a new example, that's awesome!
```

examples/README.md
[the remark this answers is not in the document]
acioc: Fair enough.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+There are a couple things you need to keep in mind, though.
```
juneb:
```suggestion
There are a couple things you need to keep in mind, though.
```
Delete this sentence

examples/src/legacy/multiple_kms_cmk_regions.py
acioc: From a correctness standpoint, I think it would be good to check them (or at least note that "in production, we would want to further verify the ARNs are valid" or something similar in case someone copy/pastes part of the example and forgets to check.

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+This example shows how to use the streaming encrypt and decrypt APIs when working with files.
+
+For the purposes of this example, we demonstrate using AWS KMS,
```
juneb:
```suggestion
This example uses an AWS KMS CMK,
```

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+but you can use other key management options with the AWS Encryption SDK.
+Look in the ``keyring`` and ``master_key_provider`` directories
+for examples that demonstrate how to use other key management configurations.
```
juneb: For X, see Y.
For examples that demonstrate how to use other key management configurations, see the ``keyring`` and ``master_key_provider`` directories

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+
+import aws_encryption_sdk
+from aws_encryption_sdk.keyrings.aws_kms import KmsKeyring
+
+
+def run(aws_kms_cmk_arn, source_plaintext_filename):
+    # type: (str, str) -> None
+    """Demonstrate an encrypt/decrypt cycle using the streaming encrypt/decrypt APIs to work with files.
+
+    :param str aws_kms_cmk_arn: AWS KMS CMK ARN to use to protect data keys
```
juneb:
```suggestion
    :param str aws_kms_cmk_arn: The ARN of an AWS KMS CMK that protects data keys
```

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    :param str source_plaintext_filename: Path to plaintext file to encrypt
+    """
+    # We assume that you can also write in the directory containing the plaintext file,
```
juneb: s/write in/write to

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    # so that is where we will put all of the results.
+    ciphertext_filename = source_plaintext_filename + ".encrypted"
+    decrypted_filename = ciphertext_filename + ".decrypted"
+
+    # Prepare your encryption context.
```
juneb: Link here?

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    encryption_context = {
+        "encryption": "context",
+        "is not": "secret",
+        "but adds": "useful metadata",
+        "that can help you": "be confident that",
+        "the data you are handling": "is what you think it is",
```
juneb: cute! :)

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    # Create the keyring that determines how your keys are protected.
```
juneb: s/keys/data keys

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    keyring = KmsKeyring(generator_key_id=aws_kms_cmk_arn)
+
+    # Open the files you want to work with.
+    with open(source_plaintext_filename, "rb") as plaintext, open(ciphertext_filename, "wb") as ciphertext:
+        # The streaming API provides you with a context manager
+        # that you can read from similar to how you would read from a file.
```
juneb:
```suggestion
        # The streaming API provides a context manager.
        # You can read from it just as you read from a file.
```

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+        with aws_encryption_sdk.stream(
+            mode="encrypt", source=plaintext, encryption_context=encryption_context, keyring=keyring
+        ) as encryptor:
+            # Iterate through the chunks in the context manager
```
juneb: Are "chunks" the elements in a stream? That makes me queasy.

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+            # and write the results to the ciphertext.
+            for chunk in encryptor:
+                ciphertext.write(chunk)
+
+    # Verify that the ciphertext and plaintext are different.
```
juneb: Kind of a low bar ...

examples/src/legacy/multiple_kms_cmk_regions.py
mattsb42-aws: I think that the right thing there will be to include an example of interacting with the keyring trace and using that to verify that the CMKs that you expected to encrypt did actually encrypt and that the one you expected to generate did actually generate.

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    assert not filecmp.cmp(source_plaintext_filename, ciphertext_filename)
+
+    # Open the files you want to work with.
+    with open(ciphertext_filename, "rb") as ciphertext, open(decrypted_filename, "wb") as decrypted:
+        # Decrypt your encrypted data.
+        #
+        # We do not need to specify the encryption context on decrypt
+        # because the header message includes the encryption context.
+        with aws_encryption_sdk.stream(mode="decrypt", source=ciphertext, keyring=keyring) as decryptor:
+            # One benefit of using the streaming API is that
```
juneb: Move this up to the streaming intro.

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+            # we can check the encryption context in the header before we start decrypting.
+            #
+            # Verify that the encryption context used in the decrypt operation matches what you expect.
```
juneb:
```suggestion
            # Verify that the encryption context used in the decrypt operation includes the encryption context that you specified when encrypting.           
```

Don't we have the usual "includes, not matches" because of the signing key?

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+            # The AWS Encryption SDK can add pairs, so don't require an exact match.
+            #
+            # In production, always use a meaningful encryption context.
+            assert set(encryption_context.items()) <= set(decryptor.header.encryption_context.items())
+
+            # Now that we are confident that the message is what we think it should be,
```
juneb:
```suggestion
            # Now that we are more confident that we will decrypt the right message,
```

examples/src/file_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+            # we can start decrypting.
+            for chunk in decryptor:
+                decrypted.write(chunk)
+
+    # Verify that the "cycled" (encrypted then decrypted) plaintext
+    # is identical to the original plaintext.
```
juneb:
```suggestion
    # Verify that the decrypted plaintext 
    # is identical to the original plaintext.
```

examples/README.md
mattsb42-aws: Not yet but they will be. I'll do a pass through the various docs when I do that, but in thinking through what to put in the migration guide I decided that providing 1:1 examples was the best way to show how to migrate.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Here you can find some examples that show you
+how to use the AWS Encryption SDK.
+We demonstrate how to use the high-level APIs
+as well as how to set up some common configuration patterns.
+
+## APIs
+
+The AWS Encryption SDK provides two high-level APIS:
+one-shot APIs that process the entire operation in memory
```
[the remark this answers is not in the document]
mattsb42-aws: That's a good point. Any ideas on what we could use instead? What I was trying to communicate is "APIs for doing everything in one operation" as opposed to "streaming APIs that require more infrastructure around them."

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+and streaming APIs.
+
+You can find examples that demonstrate these APIs
+in the [`examples/src/`](./src) directory root.
```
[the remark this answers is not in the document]
mattsb42-aws: I don't think that quite captures what I was going for. When I see "foo root directory", that implies "foo" lives somewhere underneath a root point and I want to start at "foo" and go up to its parent root. What I meant by "foo directory root" is "the files directly in the foo directory".

examples/README.md
mattsb42-aws: Higher-level than keyrings.

examples/src/file_streaming_defaults.py
mattsb42-aws: Yes. I'll update the words to match that

examples/README.md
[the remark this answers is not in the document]
mattsb42-aws: Per offline conversation, going with "one-step".

examples/src/file_streaming_defaults.py
mattsb42-aws: Doesn't have to be "chunk". The "standard" term (and what happens in more file readers) is "line", but that doesn't really apply here. I'll find a different word.

examples/src/file_streaming_defaults.py
mattsb42-aws: It is, but I'm not sure what else we can really make a concrete statement about to demonstrate that "yes, this is encrypted" without getting really complicated.

examples/src/legacy/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""
+Legacy examples.
+
+These are any examples that were already defined
+before we started revamping our examples.
+We are keeping them around for anyone who needs them as reference material,
+but we recommend looking at the newer examples
+that should provide a clearer picture of how to use this library.
+"""
```
juneb:
```suggestion
We keep these older examples as reference material, but we recommend that you use the new examples. They new examples reflect our current guidance for using the library.
"""
```

examples/src/legacy/basic_encryption.py
```diff
@@ -1,29 +1,19 @@
...
-#
-# http://aws.amazon.com/apache2.0/
-#
-# or in the "license" file accompanying this file. This file is
-# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
-# ANY KIND, either express or implied. See the License for the specific
-# language governing permissions and limitations under the License.
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
 """Example showing basic encryption and decryption of a value already in memory."""
```
juneb:
```suggestion
"""Example showing how to encrypt and decrypt a value in memory."""
```
Removing "basic" because I'm not sure how "basic encryption" differs from anything else.

examples/src/legacy/basic_encryption.py
```diff
@@ -1,29 +1,19 @@
...
 import aws_encryption_sdk
 
 
-def cycle_string(key_arn, source_plaintext, botocore_session=None):
+def run(aws_kms_cmk, source_plaintext, botocore_session=None):
```
juneb: Everyone else probably knows, but why set botocore_session to none here?

examples/src/legacy/basic_encryption.py
```diff
@@ -1,29 +1,19 @@
...
     """Encrypts and then decrypts a string under a KMS customer master key (CMK).
 
-    :param str key_arn: Amazon Resource Name (ARN) of the KMS CMK
+    :param str aws_kms_cmk: Amazon Resource Name (ARN) of the KMS CMK
```
juneb:
```suggestion
    :param str aws_kms_cmk: Amazon Resource Name (ARN) of the AWS KMS CMK
```

examples/README.md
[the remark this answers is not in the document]
mattsb42-aws: Sure.
```suggestion
in the [`examples/src/`](./src) directory.
```

examples/README.md
```diff
@@ -0,0 +1,91 @@
...
+1. Each example file MUST contain a function called `run` that runs the example.
+1. If your `run` function needs any of the following inputs,
+    the parameters MUST have the following names:
+    * `aws_kms_cmk` (`str`) : A single AWS KMS CMK ARN.
+        * NOTE: You can assume that automatically discovered credentials have
+            `kms:GenerateDataKey`, `kms:Encrypt`, and `kms:Decrypt` permissions on this CMK.
+    * `aws_kms_generator_cmk` (`str`) : A single AWS KMS CMK ARN to use as a generator key.
+        * NOTE: You can assume that automatically discovered credentials have
+            `kms:GenerateDataKey`, `kms:Encrypt`, and `kms:Decrypt` permissions on this CMK.
+    * `aws_kms_child_cmks` (`List[str]`) : A list of AWS KMS CMK ARNs to use as child keys.
```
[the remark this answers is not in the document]
mattsb42-aws: Hmm, I had missed that. I'm going to go ahead and change the examples input parameter name. We should also change this in the keyring itself, but that is out of scope for this PR. I created #220 to address that.

examples/src/in_memory_streaming_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    # https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/concepts.html#encryption-context
+    encryption_context = {
+        "encryption": "context",
+        "is not": "secret",
+        "but adds": "useful metadata",
+        "that can help you": "be confident that",
+        "the data you are handling": "is what you think it is",
+    }
+
+    # Create the keyring that determines how your data keys are protected.
```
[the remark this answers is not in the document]
mattsb42-aws: I'll talk about that in the AWS KMS keyring examples. Keyring configuration details are out of scope for this example.

examples/src/legacy/basic_encryption.py
mattsb42-aws: I'm going to be updating the legacy examples to match the format (comments/layout/etc) of the new examples once the new ones are ready. As part of that I'm actually going to be pulling that as input. That was a case of me trying to squeeze too many examples of possible configuration changes into a single example. The better place for that will be an example that specifically focuses on how to configure the KMS MKP/keyring.

Issue #215, opened 2020-02-21
[keyring] keyring trace flag values


mattsb42-aws:
When we originally made keyring trace flag values in C[1], we used bit-shifts so that they could be combined efficiently in C data structures. When we made them in Javascript[2], we went ahead and used the same values.

As the result of further discussion, we decided that we did not have any explicit intention for these values to be serialized in any way, and as such defining them with common values was potentially misleading. This is why the specification[3] does not mention any *value* for these flags, rather simply stating that what each flag means. In Java[4], we followed through on this decision and did not define any values. The flag values in Python were originally written before we came to this decision.

In order to avoid potential confusion, we should change the enum values to something else. Just a counter or an instance of `object` is fine, though it might be nice to have something explicit.

[1] https://github.com/aws/aws-encryption-sdk-c/blob/969c71d7b48a9c48e00a3ec8bc420c245681bab9/include/aws/cryptosdk/keyring_trace.h#L56-L79
[2] https://github.com/aws/aws-encryption-sdk-javascript/blob/7dfa1ae1a4696d4829c47033058d1157d6e11b0d/modules/material-management/src/keyring_trace.ts#L51-L74
[3] https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/structures.md#flags
[4] https://github.com/aws/aws-encryption-sdk-java/blob/f4973ec91598d662079159a3187d142c0e82d5b3/src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java#L20-L49

PR #225, opened 2020-03-20
feat: remove specific value definition for keyring trace flags #215

mattsb42-aws:
*Issue #, if available:*
resolves: #215

*Description of changes:*

Addresses the issues discussed in #215.

I went with a custom class rather than just `object()` because it turns out that `copy.deepcopy` cannot copy bare instances of `object`. `attrs` also gives us a nicer `repr` value.

```python
>>> class Foo(Enum):
...     @attr.s
...     class FooValue(object):
...             name = attr.ib()
...     BAR = object()
...     BAZ = FooValue("BAZ")
... 
>>> Foo.BAR
<Foo.BAR: <object object at 0x10ceefc50>>
>>> Foo.BAZ
<Foo.BAZ: Foo.FooValue(name='BAZ')>
```

# Check any applicable:


src/aws_encryption_sdk/identifiers.py
```diff
@@ -333,8 +334,14 @@ class ContentAADString(Enum):
...
+    class KeyringTraceFlagValue(object):
+        """Keyring trace flags do not have defined serializable values."""
+
+        name = attr.ib()
+
+    WRAPPING_KEY_GENERATED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_GENERATED_DATA_KEY")
+    WRAPPING_KEY_ENCRYPTED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_ENCRYPTED_DATA_KEY")
+    WRAPPING_KEY_DECRYPTED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_DECRYPTED_DATA_KEY")
+    WRAPPING_KEY_SIGNED_ENC_CTX = KeyringTraceFlagValue("WRAPPING_KEY_SIGNED_ENC_CTX")
+    WRAPPING_KEY_VERIFIED_ENC_CTX = KeyringTraceFlagValue("WRAPPING_KEY_VERIFIED_ENC_CTX")
```
WesleyRosenblum: Any reason these need to be appreviated?

src/aws_encryption_sdk/identifiers.py
```diff
@@ -333,8 +334,14 @@ class ContentAADString(Enum):
...
-    WRAPPING_KEY_DECRYPTED_DATA_KEY = 1 << 2
-    WRAPPING_KEY_SIGNED_ENC_CTX = 1 << 3
-    WRAPPING_KEY_VERIFIED_ENC_CTX = 1 << 4
+    @attr.s
+    class KeyringTraceFlagValue(object):
+        """Keyring trace flags do not have defined serializable values."""
+
+        name = attr.ib()
+
+    WRAPPING_KEY_GENERATED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_GENERATED_DATA_KEY")
```
WesleyRosenblum: Do these need to all start with "WRAPPING_KEY"?

src/aws_encryption_sdk/identifiers.py
```diff
@@ -333,8 +334,14 @@ class ContentAADString(Enum):
...
+    WRAPPING_KEY_ENCRYPTED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_ENCRYPTED_DATA_KEY")
+    WRAPPING_KEY_DECRYPTED_DATA_KEY = KeyringTraceFlagValue("WRAPPING_KEY_DECRYPTED_DATA_KEY")
```
WesleyRosenblum: You could put a comment for each of these, copying the description from the spec, like https://github.com/aws/aws-encryption-sdk-java/blob/930793314224381e37a1331b685069a0bd55e6aa/src/main/java/com/amazonaws/encryptionsdk/keyrings/KeyringTraceFlag.java#L33

src/aws_encryption_sdk/identifiers.py
mattsb42-aws: I think we went with that because that's what the prevailing naming was at the time (last summer). It looks like the names in the spec are now the simpler version. I'll update to use those names instead.

src/aws_encryption_sdk/identifiers.py
mattsb42-aws: See reason for `WRAPPING_KEY_` prefix.

PR #234, opened 2020-04-01
fix: expose the KMS keyring key namespace value for public access

mattsb42-aws:
*Description of changes:*

In order for anyone to reasonably check the keyring trace, they need to be able to access the key namespace value for the KMS keyring. This exposes that for access.

# Check any applicable:


mattsb42-aws:
Note for the record: all failing checks are expected. They are all from the test vector handler workflow. The tests from that workflow actually should not be running on pull request events, but I'll figure that out in a different PR.

Issue #220, opened 2020-03-10
[keyring] AWS KMS keyring parameters have the wrong name


mattsb42-aws:
We do not refer to the additional key IDs in the AWS KMS keyring as "child" key IDs anywhere else (spec/implementation/docs/etc). We should not introduce new terminology here.

We need to rename the parameter from `child_key_ids` to `additional_key_ids`.

PR #240, opened 2020-04-07
fix: correct KMS keyring parameter name from child_key_ids to key_ids

mattsb42-aws:
*Issue #, if available:*
resolves: #220 

# Check any applicable:

### DIFF
--- a/examples/src/keyring/aws_kms/multiple_regions.py
+++ b/examples/src/keyring/aws_kms/multiple_regions.py
@@ -46,16 +46,16 @@ def run(aws_kms_generator_cmk, aws_kms_additional_cmks, source_plaintext):
     }
 
     # Create the keyring that will encrypt your data keys under all requested CMKs.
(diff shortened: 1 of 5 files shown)


mattsb42-aws:
C calls it `additional_key_ids`

https://github.com/aws/aws-encryption-sdk-c/blob/749c84feddce147fb62e2ece7de34d733ef165c2/aws-encryption-sdk-cpp/include/aws/cryptosdk/cpp/kms_keyring.h#L129

JS calls it `keyIds`

https://github.com/aws/aws-encryption-sdk-javascript/blob/201c1022257a08c82035e5cde5097fad8321f64d/modules/kms-keyring-node/src/kms_keyring_node.ts#L42

https://github.com/aws/aws-encryption-sdk-javascript/blob/201c1022257a08c82035e5cde5097fad8321f64d/modules/kms-keyring-browser/src/kms_keyring_browser.ts#L49

I'm conflicted on the naming for this. In the context of also involving a generator key ID, I think that "additional" makes more sense because "key IDs" is just too undifferentiated. But, without the context of a generator ID, "key IDs" makes more sense because they are not in addition to anything.

Issue #207, opened 2020-01-09
Migrate CI/CD to GitHub Actions


mattsb42-aws:

farleyb-amazon:
No longer planning to do this. For now we've mostly moved towards CodeBuild.

PR #243, opened 2020-04-08
chore: disable the test vector handler tests in GitHub Actions

mattsb42-aws:
*Description of changes:*

I had been attempting to have the test vector handler tests only run on push and scheduled events, but that doesn't appear to have been working as expected. For now, this will just completely disable those tests. This is a housekeeping effort to avoid the current state where every PR has ~57 checks that are expected to fail.

Three of the static analysis checks are still expected to fail, but I think that handling those is more manageable and IMO it is better to let them keep failing until we can clean that up than to disable them. Having them fail reminds us that we need to fix them.

NOTE: These tests are still running as part of the Travis suite. We cannot run them in GitHub Actions until we figure out a solution to getting AWS credentials into GitHub Actions on pull requests.

# Check any applicable:

### DIFF
--- a/.github/workflows/ci_test-vector-handler.yaml
+++ b/.github/workflows/ci_test-vector-handler.yaml
@@ -9,9 +9,9 @@ on:
 
 jobs:
   tests:
-    # Until we address the credentials problem,
-    # do not run for pull requests.
-    if: github.event != 'pull_request'
+    # Leaving this defined but disabled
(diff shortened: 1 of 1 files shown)


mattsb42-aws:
> Do you have an issue to follow up on re-enabling this?

It's still running in Travis, so we're not missing test execution. I'm considering turning this on to be under #207

Issue #246, opened 2020-04-13
[keyring] simplify raw AES keyring API


mattsb42-aws:
As I was discussing the raw AES keyring API with @juneb, it occurred to both of us that requiring the caller to specify the wrapping algorithm does not really add value here, because the only thing that changes is the key size, which we can infer from the wrapping key material.

I think that requiring the wrapping algorithm is a hold-over from `RawMasterKey`, but one of the main reasons why the wrapping algorithm was needed to AES operations for that was to provide a consistent configuration experience for AES and RSA wrapping keys. Since we have split these out into separate keyrings, I do not think this really adds value anymore.

An argument could be made that specifying the wrapping algorithm adds value because it requires the caller to explicitly state what AES key size they intend to provide, which we can check against the key material that they actually provided. While technically correct, I'm not sure that this justifies the additional API complexity.

The other extant implementations vary on how they present this:

The specification only requires the key material:

https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/raw-aes-keyring.md#initialization

C requires a separate statement of the expected key length:

https://github.com/aws/aws-encryption-sdk-c/blob/749c84feddce147fb62e2ece7de34d733ef165c2/include/aws/cryptosdk/raw_aes_keyring.h#L57-L62

Java only requires the key material:

https://github.com/aws/aws-encryption-sdk-java/blob/5cf7e0c0638eaa6752e94b6b0da9c67a29001c9e/src/main/java/com/amazonaws/encryptionsdk/keyrings/RawAesKeyring.java#L30-L32

Javascript requires a wrapping algorithm suite:

https://github.com/aws/aws-encryption-sdk-javascript/blob/201c1022257a08c82035e5cde5097fad8321f64d/modules/raw-aes-keyring-node/src/raw_aes_keyring_node.ts#L51-L56

## Option 1: Leave as-is

If we do this, we should change the validator for the wrapping algorithm suite to not allow RSA suites.

## Option 2: Derive the wrapping algorithm suite from the wrapping key material

Given that the only thing that changes between the AES wrapping algorithm suites is the key size, we can enforce that the key material is a supported length and derive the wrapping suite from the key material length.

## Option 3: Make the wrapping algorithm suite optional

We could make the wrapping algorithm suite optional, enforcing it if the caller provides it, but otherwise deriving it from the wrapping key material.

WesleyRosenblum:
I prefer Option 2, keeping the API as simple as possible.

[...]

PR #247, opened 2020-04-13
feat: remove wrapping_algorithm input parameter from RawAESKeyring

mattsb42-aws:
*Issue #, if available:*
resolves: #246

*Description of changes:*

This implements proposed solution number 2 detailed in #246, removing wrapping algorithm as an input parameter, restricting wrapping key material length to supported lengths, and deriving the wrapping algorithm suite from the wrapping key material length.

# Check any applicable:

### DIFF
--- a/examples/src/keyring/raw_aes/raw_aes.py
+++ b/examples/src/keyring/raw_aes/raw_aes.py
@@ -10,7 +10,6 @@ In this example, we use the one-step encrypt and decrypt APIs.
 import os
 
 import aws_encryption_sdk
-from aws_encryption_sdk.identifiers import WrappingAlgorithm
 from aws_encryption_sdk.keyrings.raw import RawAESKeyring
 
 
@@ -30,14 +29,10 @@ def run(source_plaintext):
         "the data you are handling": "is what you think it is",
     }
 
-    # Choose the wrapping algorithm for the keyring to use.
-    wrapping_algorithm = WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING
-
-    # Generate an AES key to use with your keyring.
-    # The key size depends on the wrapping algorithm.
+    # Generate an AES-256 key to use with your keyring.
     #
     # In practice, you should get this key from a secure key management system such as an HSM.
-    key = os.urandom(wrapping_algorithm.algorithm.kdf_input_len)
+    key = os.urandom(32)
 
     # Create the keyring that determines how your data keys are protected.
     keyring = RawAESKeyring(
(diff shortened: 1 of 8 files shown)


src/aws_encryption_sdk/keyrings/raw.py
```diff
@@ -86,11 +85,26 @@ class RawAESKeyring(Keyring):
...
+                WrappingAlgorithm.AES_192_GCM_IV12_TAG16_NO_PADDING,
+                WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING,
+            )
+        }
+
+        try:
+            self._wrapping_algorithm = key_size_to_wrapping_algorithm[len(self._wrapping_key)]
+        except KeyError:
+            raise ValueError(
+                "Invalid wrapping key length. Must be one of {}".format(key_size_to_wrapping_algorithm.keys())
```
WesleyRosenblum, line 105: what does this error message look like, would it print out "Must be one of AES_128_GCM_IV12_TAG16_NO_PADDING, AES_192_GCM_IV12_TAG16_NO_PADDING, etc" or just the key sizes?

mattsb42-aws: It would print out the key sizes.

examples/src/keyring/raw_aes/raw_aes.py
```diff
@@ -30,14 +29,10 @@ def run(source_plaintext):
...
-    # Choose the wrapping algorithm for the keyring to use.
-    wrapping_algorithm = WrappingAlgorithm.AES_256_GCM_IV12_TAG16_NO_PADDING
-
-    # Generate an AES key to use with your keyring.
-    # The key size depends on the wrapping algorithm.
+    # Generate an AES-256 key to use with your keyring.
     #
     # In practice, you should get this key from a secure key management system such as an HSM.
-    key = os.urandom(wrapping_algorithm.algorithm.kdf_input_len)
+    key = os.urandom(32)
```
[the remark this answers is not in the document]
mattsb42-aws: line 32:

> # Generate an AES-256 key to use with your keyring.

Do you think it would be clearer to say "a 256-bit AES key"?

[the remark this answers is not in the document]
WesleyRosenblum: Yeah that sounds good, or maybe "a 256-bit (32 byte) AES key"

src/aws_encryption_sdk/keyrings/raw.py
mattsb42-aws: I updated this to be deterministic (it already was for 3.6+ but not for 3.5 or 2.7) and updated the test to catch any changes to the set.

PR #256, opened 2020-04-17
fix: move calls requiring integ test environment setup from test setup to inside test body

mattsb42-aws:
*Issue #, if available:*

related to: https://github.com/pyca/cryptography/pull/5197

*Description of changes:*

In a recent batch of test refactoring, I inadvertently moved some calls that relied on integ test environment setup (namely: setting certain environment variables) into the test setup that used to be within the test body. This caused these calls to be made during pytest test discovery. This is a problem because we use pytest markers to categorize our tests and test discovery must complete before markers can be applied to filter tests to run.

We have a test (`TOXENV=nocmk`) that should have caught this, but it looks like it was no longer removing the environment variable restrictions. That is fixed now, along with a comment explaining why that is important.

# Check any applicable:

### DIFF
--- a/test/integration/test_client.py
+++ b/test/integration/test_client.py
@@ -43,7 +43,7 @@ def _generate_mkp():
 def test_encrypt_verify_user_agent_in_logs(caplog, parameter_name, value_partial):
     caplog.set_level(level=logging.DEBUG)
 
-    aws_encryption_sdk.encrypt(source=VALUES["plaintext_128"], parameter_name=value_partial())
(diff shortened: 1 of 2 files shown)


alex:
flake8 appears unhappy

mattsb42-aws:
> flake8 appears unhappy

That's expected; in the process of moving our non-integ tests from Travis to GitHub, I enabled some checks on the test vector handler sub-module that were previously disabled. They're currently expected to fail (and so not required) but enabled to remind me to fix them.

Issue #258, opened 2020-04-22
[keyring] the filtering client suppliers need to check the client region after getting it as well as before


mattsb42-aws:
Per https://github.com/awslabs/aws-encryption-sdk-specification/issues/90, the filtering client suppliers need to check the client region after getting it from the underlying supplier before returning it.

mattsb42-aws:
closing due to KMS keyring redesign

PR #261, opened 2020-04-23
fix: filtering client suppliers MUST check the client region after creation as well as before

mattsb42-aws:
*Issue #, if available:*
resolves: #258 

*Description of changes:*

Per #258, this introduces two changes to the filtering client suppliers:
1. They now accept `None` to let users filter out requests when the region is unknown
1. In addition to checking the requested region before calling down to the inner client supplier, they now also check the client returned by the client supplier before returning it. This is to accommodate the scenario where the desired region is unknown and is set in the client by the inner client supplier.

# Check any applicable:

### DIFF
--- a/test/functional/keyrings/aws_kms/test_client_suppliers.py
+++ b/test/functional/keyrings/aws_kms/test_client_suppliers.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Functional tests for ``aws_encryption_sdk.keyrings.aws_kms.client_suppliers``."""
 import pytest
+from botocore.client import BaseClient
 from botocore.config import Config
 from botocore.session import Session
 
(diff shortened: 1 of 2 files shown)


test/functional/keyrings/aws_kms/test_client_suppliers.py
```diff
@@ -114,3 +134,24 @@ def test_allow_deny_nested_supplier():
...
+
+
+def test_allow_region_block_on_response():
+    test = AllowRegionsClientSupplier(
+        allowed_regions=[None, "us-west-2"], client_supplier=AlwaysOneRegionClientSupplier(region_name="us-east-2")
+    )
+
+    # Catch the blocked region on response.
+    with pytest.raises(UnknownRegionError):
+        test(None)
```
acioc, line 157: Should we also test the positive cases? Passing None but having it be a valid region once it's checked?

mattsb42-aws: Good call. Additional tests added.

mattsb42-aws:
Discarding this change due to upcoming major rework of AWS KMS keyrings.

https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/proposals/2020-07-01_aws-kms-keyring-redesign/change.md

Issue #44, opened 2018-02-21
avoid leaving hanging object references in helper functions


mattsb42-aws:
The `encrypt` and `decrypt` helper functions in the top-level `aws_encryption_sdk` module currently return the header object from the `EncryptionStream` objects. To avoid any possibility of those objects not being released for garbage collection because of active reference, we should copy the header object and return the copy.

abdulsmapara:
@mattsb42-aws, I am interested in solving this issue.
Does the return value of ```encrypt``` function at https://github.com/aws/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/internal/crypto/encryption.py#L76 
and ```decrypt``` function at https://github.com/aws/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/internal/crypto/encryption.py#L135 need to be modified ?

mattsb42-aws:
Hi @abdulsmapara, thanks for the interest!

This is referring to the top-level single-step interfaces for encrypt[1] and decrypt[2]. The concern is that the reference to the attribute on the stream object will keep the stream object around in memory. This is not a concern for the streaming interface because in that case the caller has direct control of the stream object.

Poking around at this a bit today, actually the only thing that we would need to do here is to use `copy.deepcopy` to copy the header and keyring trace objects from the managed stream before passing them to `CryptoResult`. It looks like `attrs` is already handling all the complicated bits. :)

[1] https://github.com/aws/aws-encryption-sdk-python/blob/82e21567a7111aea99a364e19842c3fa87a081bb/src/aws_encryption_sdk/__init__.py#L84-L87
[2] https://github.com/aws/aws-encryption-sdk-python/blob/82e21567a7111aea99a364e19842c3fa87a081bb/src/aws_encryption_sdk/__init__.py#L143-L146

abdulsmapara:
I think this PR should complete the task !

PR #264, opened 2020-04-28
hanging object references in helper functions avoided

abdulsmapara:
resolves: #44 

*Description of changes:*
- Avoided hanging object references in helper functions -
   * Created a deep copy of header and keyring trace objects from the managed stream before passing them to `CryptoResult` in the `encrypt`[1] and `decrypt`[2] functions.
    * Modified the test for checking equality of header but not identity

 [1] https://github.com/aws/aws-encryption-sdk-python/blob/82e21567a7111aea99a364e19842c3fa87a081bb/src/aws_encryption_sdk/__init__.py#L84-L87
[2] https://github.com/aws/aws-encryption-sdk-python/blob/82e21567a7111aea99a364e19842c3fa87a081bb/src/aws_encryption_sdk/__init__.py#L143-L146

### DIFF
--- a/src/aws_encryption_sdk/__init__.py
+++ b/src/aws_encryption_sdk/__init__.py
@@ -15,6 +15,7 @@ from aws_encryption_sdk.streaming_client import (  # noqa
     StreamEncryptor,
 )
 from aws_encryption_sdk.structures import CryptoResult
+import copy
 
 __all__ = ("encrypt", "decrypt", "stream")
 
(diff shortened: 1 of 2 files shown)


mattsb42-aws:
It looks like there are a bunch of tests where we currently check identity (ex: `assert header is expected_header`) that will need to be changed to check the new requirement: equality but NOT identity (ex:
```python
assert header == expected_header
assert header is not expected_header
```

Also, please run `tox -e autoformat` to take care of the formatting issues that our CI is blocking on.

abdulsmapara:
@mattsb42-aws, added commits as per the comments.

PR #288, opened 2020-08-03
fix: get CI passing again

MatthewBennington:
*Issue #, if available:* N/A

*Description of changes:* Specify the type of a sentinel passed during tests.

# Check any applicable:

### DIFF
--- a/test/unit/caches/test_local.py
+++ b/test/unit/caches/test_local.py
@@ -300,7 +300,7 @@ def patch_get_single_entry(mocker):
 def test_get_encryption_materials(patch_get_single_entry):
     cache = build_lcmc()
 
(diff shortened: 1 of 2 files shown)


test/unit/caches/test_local.py
```diff
@@ -300,6 +300,8 @@ def patch_get_single_entry(mocker):
 def test_get_encryption_materials(patch_get_single_entry):
     cache = build_lcmc()
 
+    sentinel.plaintext_length = int()
```
mattsb42-aws, line 303: `sentinel` objects are unique reference objects and we should not override them. Also, just for simplicity, rather than assigning "an int", let's just use an arbitrary integer value.
```suggestion
    plaintext_length = 42
```

PR #307, opened 2020-10-28
Update dependency on cryptography (1.7.x)

farleyb-amazon:
*Description of changes:*
Changes in 1.7.x depend on behavior introduced in cryptography 2.5, so we need to correctly model our dependency.

*Additional testing:*
If I set the requirement to `~=2.4.2` we see failures due to how cryptography checks byte types. If I set the requirement to `~=2.5.0`, it passes.


farleyb-amazon:
The pylint checks are currently failing because GitHub Actions recently added support for Python 3.9, and 3.9 seems to have a bug with pylint: https://github.com/PyCQA/pylint/issues/3882

scottarc (approved):

PR #316, opened 2021-01-15
chore: Clarify StrictAwsKmsMasterKeyProvider docs

lavaleri:
*Issue #, if available:* 

*Description of changes:* We've seen some confusion over the point that you must only use key ARNs when decrypting with the Strict KMS MK Provider. Attempting to clarify that in the code and docs.

# Check any applicable:

### DIFF
--- a/src/aws_encryption_sdk/key_providers/kms.py
+++ b/src/aws_encryption_sdk/key_providers/kms.py
@@ -231,7 +231,7 @@ class StrictAwsKmsMasterKeyProvider(BaseKMSMasterKeyProvider):
     should be used for encryption and decryption. On encryption, the plaintext will be encrypted with all configured
     master keys. On decryption, it only attempts to decrypt ciphertexts that have been wrapped with a CMK that
     matches one of the configured CMK ARNs. If the ciphertext is encrypted with a master key that was not
-    explicitly configured, decryption will fail. To create a Strict Aws Kms Master Key Provider you must provide
+    explicitly configured, decryption will fail. To create a StrictAwsKmsMasterKeyProvider you must provide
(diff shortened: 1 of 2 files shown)


lavaleri:
Note from June that saying something like "configured ARN" could be confusing for customers, as they could then think that a "Configured ARN" is a thing they need to understand. Instead, consider rewording to make it explicitly "the ARN you configured the MKP with".

src/aws_encryption_sdk/key_providers/kms.py
```diff
@@ -228,9 +228,13 @@ def _new_master_key(self, key_id):
 
 class StrictAwsKmsMasterKeyProvider(BaseKMSMasterKeyProvider):
     """Strict Master Key Provider for KMS. It is configured with an explicit list of AWS KMS master keys that
-    should be used for encryption in decryption. On encryption, the plaintext will be encrypted with all configured
-    master keys. On decryption, the ciphertext will be decrypted with the first master key that can decrypt. If the
-    ciphertext is encrypted with a master key that was not explicitly configured, decryption will fail.
+    should be used for encryption and decryption. On encryption, the plaintext will be encrypted with all configured
+    master keys. On decryption, it only attempts to decrypt ciphertexts that have been wrapped with a CMK that
+    matches one of the configured CMK ARNs. If the ciphertext is encrypted with a master key that was not
+    explicitly configured, decryption will fail. To create a Strict Aws Kms Master Key Provider you must provide
```
farleyb-amazon, line 234: Minor: looks like we're referring to the symbol here, so I would say `StrictAwsKmsMasterKeyProvider`, rather than with spaces.

farleyb-amazon (approved):

farleyb-amazon (approved):

PR #321, opened 2021-03-11
chore: Add 'release' buildspecs for codebuild

farleyb-amazon:
This is a step towards a more continuous release process. If this ends up being a productive path, we can expect to pull more validation into these codebuild specs.

*Testing*:
I've run CodeBuild builds against these specs and they succeed.

### DIFF
--- a/codebuild/release/prod-release.yml
+++ b/codebuild/release/prod-release.yml
@@ -1,0 +1,28 @@
+version: 0.2
+
+env:
+  variables:
+    BRANCH: "master"
+  secrets-manager:
+    TWINE_USERNAME: PyPiAdmin:username 
+    TWINE_PASSWORD: PyPiAdmin:password
+
+phases:
+  install:
+    runtime-versions:
+      python: latest
+  build:
+    commands:
+      - pip install tox
+      - git checkout $BRANCH
+      - tox -e park
+      - tox -e release
+      - git clone https://github.com/aws-samples/busy-engineers-document-bucket.git
+      - cd busy-engineers-document-bucket/exercises/python/encryption-context-complete
+      - tox -e test
+
+
+batch:
+  fast-fail: false
+  build-list:
+    - identifier: prod_release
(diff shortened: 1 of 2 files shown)


codebuild/release/prod-release.yml
```diff
@@ -0,0 +1,28 @@
...
+      python: latest
+  build:
+    commands:
+      - pip install tox
+      - git checkout $BRANCH
+      - tox -e park
+      - tox -e release
+      - git clone https://github.com/aws-samples/busy-engineers-document-bucket.git
+      - cd busy-engineers-document-bucket/exercises/python/encryption-context-complete
+      - tox -e test
```
robin-aws, line 22: Will this definitely pull the dependency from pypi rather than resolving it locally?

codebuild/release/prod-release.yml
```diff
@@ -0,0 +1,28 @@
...
+
+
+batch:
```
robin-aws, line 25: Why do these have to be batch builds?

codebuild/release/prod-release.yml
```diff
@@ -0,0 +1,28 @@
+version: 0.2
+
+env:
+  variables:
+    BRANCH: "master"
+  secrets-manager:
+    TWINE_USERNAME: PyPiAdmin:username 
+    TWINE_PASSWORD: PyPiAdmin:password
```
robin-aws, line 8: This is nice! It's too bad this feature doesn't seem to support downloading binary secrets (AFAICT).

codebuild/release/prod-release.yml
farleyb-amazon: Yep, we would have to jump through some hoops to try to resolve the local thing.

codebuild/release/prod-release.yml
farleyb-amazon: They don't need to be, this was mainly a) to facilitate my testing, where I was kicking off builds via the console (which uses batch builds) and b) future looking if we want to eventually take advantage of batch behavior.

robin-aws (approved):

PR #325, opened 2021-03-22
chore: Fail build if code coverage is too low

farleyb-amazon:
*Description of changes:*
Remove code coverage from base python command, add a dedicated code coverage environment, as well as code build specs to run it.

If we have all test environments running code coverage and configure it to fail on low coverage, we run into issues with things like integration tests, which intentionally are not exhaustive in their coverage. Plus running coverage on every test isn't that useful without a mechanism to do something about it (hoping a human happens to look at the output doesn't count).

So, make code coverage a dedicated target, have it target unit tests, and set a high threshold.

See also: https://github.com/aws/aws-dynamodb-encryption-python/pull/153

*Testing:*

Building gives me:
```
Required test coverage of 95.0% reached. Total coverage: 98.46%
```
If I bump the threshold to 99, the build fails with:
```
FAIL Required test coverage of 99.0% not reached. Total coverage: 98.46%
```

# Check any applicable:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,6 +9,7 @@ branch = True
 
 [coverage:report]
 show_missing = True
+fail_under = 95
 
 [tool:pytest]
 log_level = DEBUG
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -57,3 +57,6 @@ batch:
       buildspec: codebuild/py38/awses_2.0.0.yml
     - identifier: py38_awses_latest
       buildspec: codebuild/py38/awses_latest.yml
+
+    - identifier: code_coverage
+      buildspec: codebuild/coverage/coverage.yml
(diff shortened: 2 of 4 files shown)


robin-aws (dismissed):

setup.cfg
```diff
@@ -9,6 +9,7 @@ branch = True
 
 [coverage:report]
 show_missing = True
+fail_under = 95
```
robin-aws, line 12: For the record I think 95% is almost a bit too high - I like to aim for 90% so we don't hit diminishing returns, given code coverage is just an imprecise measurement. But given we're already this high I'm willing to see whether we can keep it up without feeling like we're writing a lot of low-value tests.

robin-aws (approved):

PR #326, opened 2021-03-25
chore: Add validation to test release codebuild spec

farleyb-amazon:
*Description of changes:*

Now the test release validation does the same checks as the prod release (running the sample application against the new version), but targeting testpypi. To support this I've also refactored out the validation steps into a dedicated spec so both the prod and test specs can depend on it.

*Testing:*
Prod and test releases succeed (without any version uploaded, because it hasn't changed) when I give them the correct version and commit.

When I give a bad commit, we fail to checkout the branch and stop.

When I give a good commit but bad version, we detect that the authoritative version in `identifiers.py` does not match the requested version and stop.

### DIFF
--- a/codebuild/release/prod-release.yml
+++ b/codebuild/release/prod-release.yml
@@ -20,7 +20,7 @@ phases:
       - FOUND_VERSION=$(sed -n 's/__version__ = "\(.*\)"/\1/p' src/aws_encryption_sdk/identifiers.py)
       - |
         if expr ${FOUND_VERSION} != ${VERSION}; then
-          echo "identifiers.py does not contain expected version string ${VERSION}, stopping"
+          echo "identifiers.py version (${FOUND_VERSION}) does not match expected version (${VERSION}), stopping"
           exit 1;
         fi
   build:
@@ -31,8 +31,8 @@ phases:
 batch:
   fast-fail: true 
   build-graph:
-    - identifier: prod_release
-    - identifier: validate_release
+    - identifier: release_to_prod
+    - identifier: validate_prod_release
       depend-on:
         - prod_release
       buildspec: codebuild/release/validate.yml
--- a/codebuild/release/prod-release.yml
+++ b/codebuild/release/prod-release.yml
@@ -9,21 +9,33 @@ env:
 
 phases:
   install:
-    runtime-versions:
-      python: latest
-  build:
     commands:
       - pip install tox
+      - pip install --upgrade pip
+    runtime-versions:
+      python: latest
+  pre_build:
+    commands:
       - git checkout $BRANCH
+      - CURRENT_COMMIT=$(git rev-parse --short HEAD)
+      - |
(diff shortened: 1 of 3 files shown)


codebuild/release/prod-release.yml
```diff
@@ -9,21 +9,33 @@ env:
...
+      - pip install tox
+      - pip install --upgrade pip
     runtime-versions:
       python: latest
+  pre_build:
+    commands:
+      - git checkout $COMMIT_ID
+      - FOUND_VERSION=$(sed -n 's/__version__ = "\(.*\)"/\1/p' src/aws_encryption_sdk/identifiers.py)
+      - |
+        if expr ${FOUND_VERSION} != ${VERSION}; then
```
robin-aws, line 22: I'm assuming `VERSION` has to be provided manually when executing this build? Is there a good place to document a repository's available build commands and how to call them?

codebuild/release/test-release.yml
```diff
@@ -9,17 +9,35 @@ env:
...
       - tox -e park
       - tox -e test-release
 
 
 batch:
-  fast-fail: false
-  build-list:
+  fast-fail: true
+  build-graph:
     - identifier: test_release
```
robin-aws, line 35: These labels are a little confusing: in isolation I'm not sure if this one means "test the release" or "release the test artifact". Perhaps we can use phrases like "release to production" vs "release to staging"?

codebuild/release/prod-release.yml
```diff
@@ -9,21 +9,33 @@ env:
 
 phases:
   install:
+    commands:
+      - pip install tox
+      - pip install --upgrade pip
     runtime-versions:
       python: latest
+  pre_build:
```
robin-aws, line 17: What's the functional difference between `build` and `pre_build`?

robin-aws (changes requested):

codebuild/release/test-release.yml
farleyb-amazon: Makes sense, will update wording.

codebuild/release/prod-release.yml
farleyb-amazon: To be honest I haven't found a clear description of any functional differences. The docs provide some examples of what you could do in each one, but there doesn't seem to be any concrete. My grouping here is fairly arbitrary, so I'm open to suggestions.

codebuild/release/prod-release.yml
farleyb-amazon: Correct, and good question. There is a spot for environment variables, but it looks like they do require you pass a value, and I didn't want to provide a default for the version. Worst case we could add comments, but let me look a bit more to see if there are other options.

codebuild/release/prod-release.yml
robin-aws: This is fine, it looks like `pre_build/build/post_build` are really just to help organize the phases, and this seems to fit well.

robin-aws (approved):

PR #344, opened 2021-05-27
chore: Add sleeps after releasing

farleyb-amazon:
*Description of changes:*
We see transient failures in our releases because sometimes the validation cannot find the new version that we just released due to eventual consistency. Sleep a bit between the steps to give some time for things to converge.

I couldn't find a feature within CodeBuild that allows retries or sleeps, so a manual sleep was the best I could figure out.

*Testing:*
Confirmed against the test codebuild that the happy case succeeds.

Confirmed against the test codebuild that if I request a bad version of the ESDK, it retries the appropriate number of times and then fails with an error.

### DIFF
--- a/codebuild/release/prod-release.yml
+++ b/codebuild/release/prod-release.yml
@@ -27,7 +27,6 @@ phases:
     commands:
       - tox -e park
       - tox -e release
-      - echo "Sleeping 60 seconds to give new version time to become available on PyPi" && sleep 60
 
 batch:
   fast-fail: true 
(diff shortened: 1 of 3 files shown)


robin-aws:
💯  to the motivation, but unconditional sleeping like this is such an anti-pattern. It's always better to poll in a loop instead. Couldn't we just apply a little bash loop on the validate step, with sleeping only after trying and failing, instead?

robin-aws (approved):
LGTM, thanks for the improvement!

PR #364, opened 2021-10-18
chore: Deprecate 3.5 and 3.6

josecorella:
*Issue #, if available:*

*Description of changes:*
Dropping support for 3.5 and 3.6 as they are at EOL or close to

# Check any applicable:


test_vector_handlers/setup.py
```diff
@@ -45,9 +45,6 @@ def get_requirements():
         "Natural Language :: English",
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3",
```
farleyb-amazon, line 48: We should be able to leave this one in.

PR #388, opened 2021-12-15
chore: pin dependencies and enable dependabot

josecorella:
*Description of changes:*
Pins the following dependencies:
`boto3, cryptography, attrs, wrapt`
`mock, pytest, pytest-cov, pytest-mock, bandit, doc8, flake8, pylint, black, isort`

Why were those versions picked? 
I picked these versions from the last successful github actions workflow.

# Check any applicable:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "mainline-1.x"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "mainline-2.x"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,11 +2,3 @@ boto3==1.20.24
 cryptography==36.0.1
 attrs==21.2.0
 wrapt==1.13.3
-
-bandit==1.7.1
-doc8==0.10.1
-flake8==4.0.1
-pylint==2.12.2
-black==21.12b0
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-boto3>=1.10.0
-cryptography>=2.5.0
-attrs>=17.4.0
-wrapt>=1.10.11
+boto3==1.20.24
+cryptography==36.0.1
+attrs==21.2.0
+wrapt==1.13.3
```
texastony, line 4: I think that this is a bad idea. Why? Our customers now cannot use different versions of these dependencies in their environments. I recommend reverting the changes in this file.

texastony, line 4: I think that this is a bad idea. Why? Our customers now cannot use different versions of these dependencies in their environments. I recommend reverting the changes in this file.

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
...
+    target-branch: "mainline-1.x"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "mainline-2.x"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
texastony, line 20: Do we know if dependabot will make recommendations for *any* requirements file? Or will it only think about the requirements file in root directory and ignore `test/requirements.txt`?

texastony (changes requested):
Explicitly inform us that Dependabot will consider `test/requirements.txt`.

Undo your changes to `requirements.txt`. I would rather let `dependabot` tell us version XYZ has a security issue so use `>=ABC` then commit to pinning to these latest versions that our customer's may not have in their environments yet.
I favor flexible dependencies to accommodate our customers... but I could be wrong.

.github/dependabot.yml
josecorella: It'll look at any requirements file, it's the same way with the ddbec-java repo. There are two pom.xml files, one in the root and one in sdk1/

josecorella:
Closing in favor of #389

PR #397, opened 2021-12-21
chore: Pin dependencies for linters

lavaleri:
Pin dependencies for linters to minimize broken builds on old branch

# Check any applicable:

### DIFF
--- a/dev_requirements/linter-requirements.txt
+++ b/dev_requirements/linter-requirements.txt
@@ -1,0 +1,14 @@
+bandit==1.7.0
+black==21.12b0
+doc8==0.10.1
+flake8==4.0.1
+flake8-bugbear==21.11.29
+flake8-docstrings==1.6.0
+flake8-print==4.0.0
+isort==5.10.1
+pyflakes==2.4.0
+pylint==2.12.2
+readme_renderer==32.0
+seed-isort-config==2.2.0
+vulture==2.3
+sphinx==4.3.2
--- a/tox.ini
+++ b/tox.ini
@@ -149,15 +149,13 @@ commands =
 [testenv:flake8-tests]
 basepython = {[testenv:flake8]basepython}
 deps =
-    flake8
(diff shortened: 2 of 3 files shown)


dev_requirements/linter-requirements.txt
```diff
@@ -0,0 +1,14 @@
+bandit==1.7.0
```
smswz, line 1: This is a nitpick, but can we consistently use `_` or `-` for file and directory names?

tox.ini
```diff
@@ -149,15 +149,13 @@ commands =
...
-    # https://github.com/JBKahn/flake8-print/pull/30
-    flake8-print>=3.1.0
+    -rdev_requirements/linter-requirements.txt
 commands =
     flake8 \
         # Ignore F811 redefinition errors in tests (breaks with pytest-mock use)
         # E203 is not PEP8 compliant https://github.com/ambv/black#slices
         # W503 is not PEP8 compliant https://github.com/ambv/black#line-breaks--binary-operators
-        --ignore F811,E203,W503 \
+        --ignore F811,E203,W503,D \
```
smswz, line 158: What is the `D` for?

smswz (changes requested):

tox.ini
smswz: Can we document the `D` in the comments above this line?

texastony:
```suggestion
        # D is for Documentation. We do not care if Test classes and their methods have docstrings
        --ignore F811,E203,W503,D \
```

dev_requirements/linter-requirements.txt
texastony: In this repo, there is a mostly consistent policy that directories use `_` and files use `-`.

I do not see the value in this policy. 
I do not know how this came to be.
I am not going suggest we enforce it in any way. 
But since this already meets the policy, I suggest we leave it as is.

texastony (approved):

smswz (approved):

PR #440, opened 2022-03-09
chore: Add example for custom KMS client config

seebees:
Occasionally custom configuration for the AWS KMS client is needed.
This is especially important in Discovery mode,
But anytime you are dealing with more than one AWS KMS Key,
the master key providers are simpler to work with then individual master keys.

# Check any applicable:


examples/src/custom_kms_client_for_kms_provider.py
```diff
@@ -0,0 +1,74 @@
...
+            session = boto3.session.Session(botocore_session=self.config.botocore_session)
+            client = session.client(
+                'kms',
+                region_name=region_name,
+                config=self._user_agent_adding_config,
+                # Add additional custom client configuration here
+                connect_timeout=10
+            )
+            self._register_client(client, region_name)
+            self._regional_clients[region_name] = client
```
mattsb42-aws, line 36: Just a note: IMO by including this in the official examples, this implicitly blesses these private methods as part of the public API for this class. Not sure if that's an issue or not, depending on the plans for the future of the codebase. :)

seebees: Yes, it may be better to refactor this example as:

```
main_mk = StrictAwsKmsMasterKeyProvider(providerConfig)
main_mk.add_master_key(MRKAwareKMSMasterKey(anotherConfig))
```

However this does not address discovery.

alex-chew:
This will resolve #446.

texastony (dismissed):

justplaz (approved):
Reapproving after syntax fixes.

PR #492, opened 2022-08-03
refactor: reduce logging verbosity

cjkindel:
*Issue #, if available:*

*Description of changes:*
Similar logging is classified at DEBUG level. There is also no value added to a functioning implementation as the log statements just help document stepping through lines, which in my opinion is a debugging activity. There is no 'INFO' being added by these noisy log statements.

# Check any applicable:

### DIFF
--- a/src/aws_encryption_sdk/streaming_client.py
+++ b/src/aws_encryption_sdk/streaming_client.py
@@ -299,10 +299,10 @@ class _EncryptionStream(io.IOBase):
 
     def readline(self):
         """Read a chunk of the output"""
-        _LOGGER.info("reading line")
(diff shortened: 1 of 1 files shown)


texastony (approved):
Howdy @cjkindel,

Thank you for the PR. 
I agree that the ESDK-Python's logging is more verbose than needed.
And I am happy to lower that verbosity.

Much Obliged,
AWS Crypto Tools

texastony:
@cjkindel 
You will need to make another commit that conforms to [Semantic Rules.](https://cheatography.com/albelop/cheat-sheets/conventional-commits/pdf/)
I would suggest:
`refactor: reduce logging verbosity`

PR #502, opened 2022-08-26
chore(release): add api token to staging release process

josecorella:
*Description of changes:*
Uses API token instead of username and password to publish artifacts to TestPyPi. Token only has access to publish you cannot log into the account with it.

# Check any applicable:


josecorella:
this change cannot be tested in ci, i have to first merge this pr to then run the CB so that it uses the new secret

justplaz (approved):
LGTM

PR #548, opened 2023-01-23
chore: freeze py311 depedencies for pyca and test in CI

texastony:
*Issue #, if available:* https://github.com/pyca/cryptography/pull/8115

*Description of changes:* 
- Explicitly test Python 3.10 & Python 3.11 in CI.
- Update Freeze Dependencies for Python 3.7
- Add Freeze Dependencies for Python 3.11

# Check any applicable:

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -31,7 +31,7 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
-          - 3.10
+          - "3.10"
           - "3.11"
           - 3.x  # Ideally, we would skip if 3.x is 3.11
         architecture:
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -31,7 +31,9 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
-          - 3.x
+          - 3.10
+          - 3.11
+          - 3.x  # Ideally, we would skip if 3.x is 3.11
         architecture:
           - x64
           - x86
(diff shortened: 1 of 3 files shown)


texastony:
There are two checks that are expected and required,
but will never run, 
because I re-named them...
Screen Shot 2023-01-23 at 12 07 55
I will have to update the branch protection to require:
- Continuous Integration tests / upstream-py311 (nocmk)
- Continuous Integration tests / upstream-py311 (test-upstream-requirements-py311) 
- Continuous Integration tests / upstream-py37 (test-upstream-requirements-py37)
- Continuous Integration tests / upstream-py37 (nocmk)

test/upstream-requirements-py37.txt
```diff
@@ -1,26 +1,25 @@
-attrs==21.2.0
```
josecorella, line 1: @texastony, how come we are changing this file? are we just refreshing the deps here?

from the conversation over in https://github.com/pyca/cryptography/pull/8115#issuecomment-1400769056
we are doing both of their suggestions. Is there a need for both or can we just update this file?

texastony: Actually, I am some-what interested in the diff of these two files.
But yes, by updating py37.txt, we are updating the dependencies frozen there to those used when testing with python 3.7.12.

texastony: I am not against dropping 3.7 entirely, but I think that Pyca depends on it...
This way, we offer them both, and if they ever tell us we only need 311, we can drop 3.7.

@alex , can you let us know if you no longer need 3.7?

alex: We don't need py37 anymore -- we only test our downstreams against a single python version.

PR #627, opened 2024-01-30
chore: skip pyenv installation if already exists

imabhichow:
*Issue #, if available:*

*Description of changes:*

- codebuild ci is failing with `pyenv 3.12.0 already exists`, with this change ci will skip `pyenv` installation if already exists

# Check any applicable:


texastony (approved):
This is great! We could/should always has this flag in CI, but let's leave that for another day

PR #634, opened 2024-02-02
feat: Add MPL and hierarchy keyring example

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Add MPL support to ESDK-Python. This is an optional dependency that is only supported on Python 3.11+.
* Create an example using the MPL's hierarchical keyring. This is mirrored from [ESDK-Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/examples/java/com/amazonaws/crypto/examples/keyrings/hierarchical/AwsKmsHierarchicalKeyringExample.java).
* Update CI:
  * No change to coverage for Python <3.11
  * Python 3.11+ now runs 3 test suites
    * Native ESDK tests WITHOUT the MPL installed in the environment
    * Native ESDK tests WITH the MPL installed in the environment
    * Native ESDK tests AND MPL-specific tests WITH the MPL installed in the environment

# Check any applicable:


examples/src/keyrings/hierarchical_keyring.py
```diff
@@ -0,0 +1,204 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Example showing basic encryption and decryption of a value already in memory."""
+import sys
+
+import boto3
+# ignore missing MPL for pylint, but the MPL is required for this example
+# noqa pylint: disable=import-error
+from aws_cryptographic_materialproviders.keystore.client import KeyStore
```
texastony, line 9: Question: Is it possible we could reference `KeyStore` in  `aws_cryptographic_materialproviders.keystore.__init__`,
such that users could write
`from aws_cryptographic_materialproviders.keystore import KeyStore`
?

src/aws_encryption_sdk/internal/crypto/authentication.py
texastony, line 1: Question: Are these changes something we should push into ESDK-Python now?

Are they induced by a change in `pyca/cryptography` or by the MPL?

src/aws_encryption_sdk/materials_managers/mpl/cmm.py
```diff
@@ -0,0 +1,142 @@
...
+    from aws_cryptographic_materialproviders.mpl.models import (
+        AlgorithmSuiteIdESDK,
+        CommitmentPolicyESDK,
+        DecryptMaterialsInput,
+        DecryptMaterialsOutput,
+        EncryptedDataKey as MPL_EncryptedDataKey,
+        GetEncryptionMaterialsInput,
+        GetEncryptionMaterialsOutput,
+    )
+    from aws_cryptographic_materialproviders.mpl.references import ICryptographicMaterialsManager
```
texastony, line 17: Issue/Suggestion: I think I would preface all of these with `MPL_`.

The names are so similar, I wanted to go the distance to avoid confusion.

src/aws_encryption_sdk/materials_managers/mpl/cmm.py
```diff
@@ -0,0 +1,142 @@
...
+
+except ImportError:
+    pass
+
+from typing import List
+
+from aws_encryption_sdk.exceptions import AWSEncryptionSDKClientError
+from aws_encryption_sdk.identifiers import CommitmentPolicy
+from aws_encryption_sdk.materials_managers.mpl.materials import MPLEncryptionMaterials, MPLDecryptionMaterials
```
texastony, line 26: Issue/Suggestion: I would be consistent with the prefix. Always use `MPL_ `.

texastony, line 26: Question: Should this import be in the try block?

This:  `aws_encryption_sdk.materials_managers.mpl.materials`

texastony (commented):
This is a really cool PR.
It demonstrates the flexibility of Python's interpretive nature.

Any customer's using a non-interpretive Python environment will be in trouble
(Jython?),
but we have never stated support for anything but CPython, 
so this is fine.

examples/src/keyrings/hierarchical_keyring.py
lucasmcdonald3: That should possible, but it certainly crosses into the territory of questionable namespace hacking...

lucasmcdonald3: I think it's reasonable for the primary local service client class to be accessible from this module. Will update.

src/aws_encryption_sdk/internal/crypto/authentication.py
lucasmcdonald3: The code this linked to is gone, but assuming it's on the `from_key_bytes` changes:

There is no use in pushing this to ESDK-Python now.
This is from the MPL. The MPL encodes as PEM, but ESDK encodes as DER.

src/aws_encryption_sdk/materials_managers/mpl/cmm.py
lucasmcdonald3: Not in this case. This particular module will require the MPL to be installed to import it.

lucasmcdonald3: These names are a little more nuanced, let me give some more thought to this.
These are classes for the MPL materials that conform to native interfaces.
These internally talk to a `DecryptionMaterials as MPL_DecryptionMaterials` and a `EncryptionMaterials as MPL_EncryptionMaterials` which are actually from the MPL.
If `MPL_` prefix prefixes anything "from the MPL", these shouldn't have the prefix

lucasmcdonald3: I'm currently at `EncryptionMaterialsFromMPL` and `DecryptionMaterialsFromMPL` but welcome other suggestions

texastony: I would follow the universal construct of Namespaces before Class names.
i.e: com.amazon.encryption.Encrypt vs Encrypt from com.amazon.Encryption.

I think these are constructs of this library (ESDK-Python)
that wrap constructs from the MPL.

Hence your statement:
> If `MPL_` prefix prefixes anything "from the MPL", these shouldn't have the prefix

However, I disagree with your conclusion.
Just b/c the construct is defined outside of the MPL
(in fact, it is defined in the ESDK-Python's `aws_encryption_sdk.materials_managers.mpl.materials` module)
does not mean it cannot be prefixed with `MPL_`.

I think it is acceptable to prefix anything with `MPL_` if it helps clarify 
that the construct is being introduced/referenced to work with the MPL 
as compared to the existing ESDK-Python.

lucasmcdonald3: > I think it is acceptable to prefix anything with MPL_ if it helps clarify that the construct is being introduced/referenced to work with the MPL

Agreed but the problem is that there are both
`aws_cryptographic_materialproviders.mpl.models.EncryptionMaterials` that could reasonably called `MPL_EncryptionMaterials`
and
`aws_encryption _sdk.materialsmanagers.mpl.materials.SomeClassNameThatExtends(Native_EncryptionMaterials)`, where `SomeClassNameThatExtends` is currently called `EncryptionMaterialsFromMPL` but could also be called `MPL_EncryptionMaterials`.
I welcome suggestions here since I'm not happy with the current state, but if nobody has alternatives I'd be fine leaving this as-is.

seebees (changes requested):
This is looking good!
Going over the testing is the thing. I would like to go over the test vector plans.

seebees (changes requested):
We need to deal with Encryption Context on decrypt and the required encryption context CMM.

src/aws_encryption_sdk/materials_managers/mpl/cmm.py
texastony: You could do something like `Wrapped_MPL`,
to keep everything distinctive... but I agree... at the end of the day, this all pretty confusing...

Maybe we need to do more surgery/refactoring to the ESDK-Python
so that the MPL's materials are a valid extension of the native materials?

That could cut down on the wrapping.

README.rst
```diff
@@ -35,7 +35,7 @@ Required Prerequisites
 ======================
 
 * Python 3.7+
-* cryptography >= 2.5.0
+* cryptography >= 3.4.6
```
seebees, line 38: Is this a breaking change?
What is going to be the version bump we are pushing here?

requirements.txt
```diff
@@ -1,4 +1,4 @@
 boto3>=1.10.0
-cryptography>=3.4.0
+cryptography>=3.4.6
```
seebees, line 2: Because I don't have it to hand: What version is the one that starts to require Rust?

setup.py
```diff
@@ -39,6 +39,13 @@ def get_requirements():
     keywords="aws-encryption-sdk aws kms encryption",
     license="Apache License 2.0",
     install_requires=get_requirements(),
+    # pylint: disable=fixme
+    # TODO: Point at MPL main branch once Python MPL is merged into main.
+    extras_require={
+        "MPL": ["aws-cryptographic-material-providers @" \
+                "git+https://github.com/aws/aws-cryptographic-material-providers-library.git@" \
+                "lucmcdon/python-mpl#subdirectory=AwsCryptographicMaterialProviders/runtimes/python"],
```
seebees, line 47: This should point to a specific version yes?

src/aws_encryption_sdk/materials_managers/mpl/cmm.py
```diff
@@ -0,0 +1,147 @@
...
+from aws_encryption_sdk.materials_managers.base import CryptoMaterialsManager
+from aws_encryption_sdk.materials_managers.mpl.materials import DecryptionMaterialsFromMPL, EncryptionMaterialsFromMPL
+from aws_encryption_sdk.structures import EncryptedDataKey as Native_EncryptedDataKey
+
+
+class CryptoMaterialsManagerFromMPL(CryptoMaterialsManager):
+    """
+    In instances where encryption materials are provided by an implementation of the MPL's
+    `aws_cryptographic_materialproviders.mpl.references.MPL_ICryptographicMaterialsManager`,
+    this maps the ESDK CMM interfaces to the MPL CMM.
```
seebees, line 39:
```suggestion
    this maps the ESDK-Python CMM interfaces to the MPL CMM.
```

src/aws_encryption_sdk/materials_managers/mpl/materials.py
```diff
@@ -0,0 +1,138 @@
...
+)
+from aws_encryption_sdk.structures import DataKey, EncryptedDataKey as Native_EncryptedDataKey, MasterKeyInfo
+
+
+def _mpl_algorithm_id_to_native_algorithm_id(mpl_algorithm_id: str) -> int:
+    # MPL algorithm suite ID == hex(native algorithm suite ID)
+    return int(mpl_algorithm_id, 16)
+
+
+class EncryptionMaterialsFromMPL(Native_EncryptionMaterials):
```
seebees, line 30: What about required EC and the like?

test/upstream.md
seebees, line 1: nice thanks

requirements.txt
lucasmcdonald3: This change is already on master: https://github.com/aws/aws-encryption-sdk-python/blob/master/requirements.txt
I'll do some git to fix this diff

setup.py
lucasmcdonald3: Yeah... I'll update the comment here

README.rst
lucasmcdonald3: Again, this is on master already: https://github.com/aws/aws-encryption-sdk-python/blob/master/README.rst?plain=1#L38
will fix

src/aws_encryption_sdk/materials_managers/mpl/materials.py
lucasmcdonald3: I'll plan to leave all required EC "stuff" in https://github.com/aws/aws-encryption-sdk-python/pull/645

requirements.txt
lucasmcdonald3: Done, needed to merge master into `mpl-reviewed` feature branch

README.rst
lucasmcdonald3: Done, needed to merge master into `mpl-reviewed` feature branch

setup.py
lucasmcdonald3: Done

ajewellamz (approved):
Looks good.

seebees (approved):
Looks like everything I cared about is now in #645.

PR #645, opened 2024-02-29
feat: Required encryption context CMM

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Support required encryption context CMM
* Add example using required encryption context CMM

# Check any applicable:


src/aws_encryption_sdk/internal/formatting/serialize.py
```diff
@@ -257,13 +289,21 @@ def serialize_header_auth(version, algorithm, header, data_encryption_key, signe
...
     :returns: Serialized header authentication data
     :rtype: bytes
     """
     if version == SerializationVersion.V1:
         return _serialize_header_auth_v1(algorithm, header, data_encryption_key, signer)
     elif version == SerializationVersion.V2:
-        return _serialize_header_auth_v2(algorithm, header, data_encryption_key, signer)
+        return _serialize_header_auth_v2(
+            algorithm, header, data_encryption_key, signer, required_ec_bytes
+        )
```
seebees, line 306: I don't think that we care about the diff between v2 and v1 for this.

test/unit/test_streaming_client_stream_decryptor.py
seebees, line 1: This is a good start.
Here are the basic tests that we have done on the Dafny side
https://github.com/aws/aws-encryption-sdk-dafny/blob/mainline/AwsEncryptionSDK/dafny/AwsEncryptionSdk/test/TestReproducedEncContext.dfy

src/aws_encryption_sdk/materials_managers/mpl/materials.py
```diff
@@ -136,3 +143,15 @@ def data_key(self) -> DataKey:
...
+
+    @property
+    def encryption_context(self) -> Dict[str, str]:
+        """Materials' encryption context."""
+        return self.mpl_materials.encryption_context
+
+    @property
+    # Pylint thinks this name is too long, but it's the best descriptor for this...
+    # pylint: disable=invalid-name
+    def required_encryption_context_keys(self) -> bytes:
```
seebees, line 155: I think that this is `reproducedEncryptionContext`. The two sides are not symmetrical.

lucasmcdonald3: This is how it's called in the MPL `DecryptionMaterials`: https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/structures.smithy#L79

It is true that the `DecryptionMaterialsRequest` takes in a `reproducedEncryptionContext` argument: https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/cmms.smithy#L108

I believe the `reproducedEncryptionContext` from the request is passed into the Dafny layer, where it is validated, then the "response" here includes this as output.

examples/src/keyrings/hierarchical_keyring.py
```diff
@@ -25,6 +55,7 @@
 
 from .example_branch_key_id_supplier import ExampleBranchKeyIdSupplier
 
+# TODO-MPL: Remove this as part of removing PYTHONPATH hacks
```
lucasmcdonald3, line 58:
```suggestion
# TODO-MPL: Remove this as part of removing PYTHONPATH hacks.
```

examples/src/keyrings/hierarchical_keyring.py
```diff
@@ -1,6 +1,36 @@
...
+the tenant's data. You can either statically configure a single branch key to ensure you are
+restricting access to a single tenant, or you can implement an interface that selects the Branch
+Key based on the Encryption Context.
+
+This example demonstrates configuring a Hierarchical Keyring with a Branch Key ID Supplier to
+encrypt and decrypt data for two separate tenants.
+
+This example requires access to the DDB Table where you are storing the Branch Keys. This
+table must be configured with the following primary key configuration: - Partition key is named
+"partition_key" with type (S) - Sort key is named "sort_key" with type (S)
```
lucasmcdonald3, line 29:
```suggestion
"partition_key" with type (S) - Sort key is named "sort_key" with type (S).
```

src/aws_encryption_sdk/__init__.py
```diff
@@ -185,6 +185,9 @@ def decrypt(self, **kwargs):
                 If source_length is not provided and read() is called, will attempt to seek()
                 to the end of the stream and tell() to find the length of source data.
 
+        :param dict encryption_context: Dictionary defining encryption context to validate
+            on decrypt. This is ONLY validated on decrypt if using the required encryption
+            context CMM from the aws-cryptographic-materialproviders library.
```
seebees, line 190: This should be validated by the default CMM as well.

src/aws_encryption_sdk/internal/formatting/serialize.py
```diff
@@ -227,16 +258,35 @@ def _serialize_header_auth_v2(algorithm, header, data_encryption_key, signer=Non
...
+    else:
+        header_auth = encrypt(
+            algorithm=algorithm,
+            key=data_encryption_key,
+            plaintext=b"",
+            # The AAD MUST be the concatenation of the serialized message header body and the serialization
+            # of encryption context to only authenticate. The encryption context to only authenticate MUST
+            # be the encryption context in the encryption materials filtered to only contain key value
+            # pairs listed in the encryption material's required encryption context keys serialized
+            # according to the encryption context serialization specification.
```
seebees, line 286: This looks like a copy from the spec? I would suggest that you just use Duvet

src/aws_encryption_sdk/internal/formatting/serialize.py
```diff
@@ -257,13 +314,22 @@ def serialize_header_auth(version, algorithm, header, data_encryption_key, signe
     :param bytes data_encryption_key: Data key with which to encrypt message
     :param signer: Cryptographic signer object (optional)
     :type signer: aws_encryption_sdk.Signer
+    :param required_encryption_context_bytes: Serialized encryption context items
+        for all items whose keys are in the required_encryption_context list.
+        This is ONLY processed if using the aws-cryptographic-materialproviders library
+        AND its required encryption context CMM. (optional)
```
seebees, line 320: Maybe I'm not following, but the Default CMM should do this. Getting required EC only happens with the required EC CMM _on encrypt_.

seebees (changes requested):
The default CMM, or really any CMM should be able to handle reproduced EC. We may need to re-think how we want to expose the Default CMM in this new world.

src/aws_encryption_sdk/internal/formatting/serialize.py
lucasmcdonald3: I actually copied this from ESDK-Java...
This repo doesn't have Duvet set up
I may take a look at setting it up

src/aws_encryption_sdk/__init__.py
lucasmcdonald3: Note: "default CMM" is ONLY the MPL default CMM. The native ESDK-Python default CMM will NOT validate this.
All native ESDK-Python CMMs will be marked as deprecated as part of the release for MPL integration.
Deprecated components will not support new MPL features (like reproduced encryption context on decrypt).
(We may add support for reproduced EC to native default CMM at some point, but not for MPL integration launch.)

src/aws_encryption_sdk/internal/formatting/serialize.py
lucasmcdonald3: Fixed; this was already the case in the code, but not in the documentation.

examples/src/keyrings/hierarchical_keyring.py
```diff
@@ -25,6 +55,7 @@
 
 from .example_branch_key_id_supplier import ExampleBranchKeyIdSupplier
 
+# TODO-MPL: Remove this as part of removing PYTHONPATH hacks.
```
ajewellamz, line 58: I assume this TODO is not yet TO BE DONE?

lucasmcdonald3: This is still TODO. I have to:
* Update Dafny compiler to support outer-module
* Update Smithy-Dafny to reference outer-module
* Update MPL to use outer-module
I'm tracking that work separately.

ajewellamz (approved):
LGTM

seebees (approved):
LGTM

PR #650, opened 2024-03-07
feat(test_vector_handlers): TestVectors test with MPL constructs

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

* Modify test vector suite for all supported Python versions:
  * Add test decrypting generated test vectors from the most recent successful run of ESDK-Dafny "Daily CI": [current most recent](https://github.com/aws/aws-encryption-sdk-dafny/actions/runs/8328902651)
    * "most recent" will update daily, so the vectors used will update daily
  * Add test decrypting required encryption context CMM manifest: [link](https://github.com/aws/aws-encryption-sdk-dafny/raw/mainline/AwsEncryptionSDK/runtimes/net/TestVectorsNative/TestVectors/resources/v4-Net-4.0.1.zip)
    * For Python <3.10, MPL is not supported, and required encryption context CMM is not supported. However, this manifest also has "Default" CMM scenarios. As part of these changes, test vector runners without the MPL installed will filter out scenarios with the required encryption context CMM, and only test the "Default" CMM scenarios.
  * Add test generating decrypt vectors with master key constructs, then decrypting them with Python master key constructs and the JS decrypt oracle.
  * Add test encrypting vectors with master key constructs.
  * Remove running `awses_local`; this test is now redundant by the additions above.
     * This consisted of 2 tests: 1) "test encrypt"; 2) "test generate decrypt vectors, then decrypt those vectors". These are now covered by the above.
* Add test vectors if the MPL is installed:
  * Add test generating decrypt vectors with keyring constructs, then decrypting them with Python keyrings, Python master keys, and the JS decrypt oracle.
  * Add test decrypting the master key-generated vectors with keyrings.
  * Add test encrypting vectors with keyring constructs.

*Out of Scope*

* Hierarchy keyring test vectors. Tracked in #674.
* Decrypt with NET. Will be tracked separately. This requires publishing .NET MPL TestVectors and merging in https://github.com/aws/aws-encryption-sdk-dafny/pull/649.

# Check any applicable:


src/aws_encryption_sdk/internal/formatting/deserialize.py
```diff
@@ -475,7 +475,7 @@ def deserialize_frame(stream, header, verifier=None):
     frame_data["iv"] = frame_iv
     if final_frame is True:
         (content_length,) = unpack_values(">I", stream, verifier)
-        if content_length >= header.frame_length:
+        if content_length > header.frame_length:
```
lucasmcdonald3, line 478: Is this concerning at all? Some tests were failing before this change, and succeeded after.

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
```diff
@@ -0,0 +1,182 @@
...
+
+        }
+        if self.padding_algorithm is not None and self.padding_algorithm != "":
+            input_kwargs["padding-algorithm"] = self.padding_algorithm
+        if self.padding_hash is not None:
+            input_kwargs["padding-hash"] = self.padding_hash
+
+        if input_kwargs["type"] == "raw" \
+                and input_kwargs["encryption-algorithm"] == "rsa":
+            # Weird hack #1:
```
lucasmcdonald3, line 85: Weird hack #1 is also present in Java. https://github.com/aws/aws-encryption-sdk-java/blob/master/src/test/java/com/amazonaws/encryptionsdk/TestVectorGenerator.java#L414-L420

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
```diff
@@ -0,0 +1,182 @@
...
+            GetKeyDescriptionInput(json=encoded_json)
+        )
+
+        keyring: IKeyring = keyvectors.create_test_vector_keyring(
+            TestVectorKeyringInput(
+                key_description=output.key_description
+            )
+        )
+
+        # Weird hack #2:
```
lucasmcdonald3, line 117: Weird hack #2 is NOT present in Java.
Java test vectors use a modified keys.json file from other manifests: https://github.com/aws/aws-encryption-sdk-java/blob/master/src/test/resources/keys.json#L45
This is changed from `"key-id": "rsa-4096-private"` to `"key-id": "rsa-4096"`; similarly on the public key.
This allows the `edk.keyProviderInfo == keyring.keyName` check to pass, and the keyring "works".
But (afaik) all of these vector suites provide their own `keys.json` that still have `key-ids` that specify `private`/`public`, so Java would not work with those vector suites.

This code should probably be in the TestVectors Dafny. It's only here because I debugged this in Python, and want some feedback before changing the Dafny.

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
seebees: Does this mean we have a simplified tests. Eg it specifies the private key to both encrypt and decrypt?

seebees (commented):
So we generate with master key, and then decrypt with JS and master key.
How do we test keyring? Does the master key target do both?

test_vector_handlers/src/awses_test_vectors/internal/mpl/tampering_mpl_materials.py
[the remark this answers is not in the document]
lucasmcdonald3: If it isn't already, I think that's a good idea

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
lucasmcdonald3: I'd describe it as "coordinating" rather than fighting :)
The primary evidence for this is that the test vectors pass with this change

I think the plan of action here SHOULD be some combination of 1) rename key manifests going forward so the key name of a public/private key pair is the same, 2) add a "legacy" check in TestVectors to align the key name of a public/private key pair

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
lucasmcdonald3: The `decrypt-generate` (i.e. `encrypt`) vector manifest specifies a "private" key, but it really should be grabbing the public key.

test_vector_handlers/src/awses_test_vectors/internal/mpl/tampering_mpl_materials.py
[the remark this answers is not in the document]
lucasmcdonald3: Chatted with Jose, it's a good idea to put this in the test vector framework.

test_vector_handlers/src/awses_test_vectors/manifests/mpl_keyring.py
seebees: Is this plan of action going to happen here or is this happening somewhere else?

test_vector_handlers/src/awses_test_vectors/internal/mpl/tampering_mpl_materials.py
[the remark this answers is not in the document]
seebees: Can I resolve this? If this is done, just resolve the conversation plz :)

setup.py
```diff
@@ -40,7 +40,8 @@ def get_requirements():
     license="Apache License 2.0",
     install_requires=get_requirements(),
     # pylint: disable=fixme
-    # TODO: Point at MPL main branch once Python MPL is merged into main.
+    # TODO-MPL: Point at PyPI once MPL is released.
+    # This blocks releasing ESDK-Python MPL integration.
```
seebees, line 44: Just wondering do we have a way to enforce this?

test/unit/test_streaming_client_stream_decryptor.py
```diff
@@ -364,142 +365,6 @@ def test_GIVEN_verification_key_AND_has_mpl_AND_has_MPLCMM_WHEN_read_header_THEN
             algorithm=self.mock_header.algorithm, encoded_point=mock_b64encoding()
         )
 
-    @patch("aws_encryption_sdk.streaming_client.derive_data_encryption_key")
-    @patch("aws_encryption_sdk.streaming_client.DecryptionMaterialsRequest")
-    @patch("aws_encryption_sdk.streaming_client.Verifier")
-    # Given: no MPL
-    @pytest.mark.skipif(HAS_MPL, reason="Test should only be executed without MPL in installation")
-    def test_GIVEN_decrypt_config_has_ec_WHEN_read_header_THEN_calls_decrypt_materials_with_reproduced_ec(
-        self,
```
seebees, line 244: Just wondering why we are deleting these?

seebees (approved):
Looks good! So many commits!

setup.py
lucasmcdonald3: Yes! PyPI will not let you publish a package that uses GitHub directly. We MUST refer to the published MPL to publish the new ESDK-Python.

lucasmcdonald3: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#

test/unit/test_streaming_client_stream_decryptor.py
lucasmcdonald3: This might be a diff issue... let me double check

lucasmcdonald3: Ah! This was a merge conflict issue. I committed this test when I shouldn't have.
The behavior from this test is no longer valid.
This test would assert that encryption context on decrypt works for native (ESDK-Python) CMMs.
We decided that encryption context on decrypt would not be supported in this [thread](https://github.com/aws/aws-encryption-sdk-python/pull/645#discussion_r1569435020).
The new behavior validating encryption context on decrypt raises an exception is present [here](https://github.com/aws/aws-encryption-sdk-python/blob/mpl-reviewed/test/unit/test_streaming_client_stream_decryptor.py#L306).

PR #661, opened 2024-04-24
chore: added examples for raw rsa and raw aes keyrings

RitvikKapila:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:


examples/src/keyrings/raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,191 @@
...
+decrypts the data key using the private key.
+
+This example creates a Raw RSA Keyring and then encrypts a custom input EXAMPLE_DATA
+with an encryption context. This example also includes some sanity checks for demonstration:
+1. Ciphertext and plaintext data are not the same
+2. Encryption context is correct in the decrypted message header
+3. Decrypted plaintext value matches EXAMPLE_DATA
+4. After verifying that the encrypt and decrypt works, this example also demonstrates
+that the original ciphertext should not be decrypted using a new Raw RSA keyring generated by
+another user, let's say Bob (Points 9 and 10).
```
lucasmcdonald3, line 22:
```suggestion
4. The original ciphertext is not decryptable using a keyring with a different RSA key pair
```

examples/src/keyrings/raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,191 @@
...
+
+# TODO-MPL: Remove this as part of removing PYTHONPATH hacks.
+MODULE_ROOT_DIR = '/'.join(__file__.split("/")[:-1])
+
+sys.path.append(MODULE_ROOT_DIR)
+
+EXAMPLE_DATA: bytes = b"Hello World"
+
+
+def generate_rsa_keyring():
```
lucasmcdonald3, line 61: Can we refactor this so the logic is

```
if user has provided private/public key PEMs:
    load those PEMs
otherwise:
    generate key pair, write PEM to file
load keyring from PEMs

```

We should allow users to bring their own PEMs to this example.
Most users who are interested in a raw RSA keyring will already have PEM-encoded files that they want to provide.
We should let them feed those keys into this example.
But, this also gives them some sample code that they can copy-paste to load their keys in their own code, rather than having to write load-from-file code themselves.

I see you're doing what we did in [Java-ESDK](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/examples/java/com/amazonaws/crypto/examples/keyrings/RawRsaKeyringExample.java), but that example is not great.
The [NET-ESDK](https://github.com/aws/aws-encryption-sdk-dafny/blob/mainline/AwsEncryptionSDK/runtimes/net/Examples/Keyring/RawRSAKeyringExample.cs), we also doesn't do a good job of this; it writes stuff to files, but doesn't let users bring their own files out of the box.
The model here should be [Java-DBESDK](https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/Examples/runtimes/java/DynamoDbEncryption/src/main/java/software/amazon/cryptography/examples/keyring/RawRsaKeyringExample.java).

Feel free to push back if this is too much and we can find a middle ground

examples/test/keyrings/test_i_raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,58 @@
...
+    Here user provides only the public key. The program should throw an Assertion error
+    as this example requires the user to either provide both private and public keys to
+    test both encryption and decryption, or not provide any keys and the example generates both
+    """
+    user_public_key, user_private_key = generate_rsa_keys()
+    try:
+        encrypt_and_decrypt_with_keyring(public_key=user_public_key)
+
+        raise AssertionError("encrypt_and_decrypt_with_keyring should raise an error")
+    except AssertionError:
```
lucasmcdonald3, line 41: Let's do ValueError

examples/src/keyrings/raw_aes_keyring_example.py
```diff
@@ -0,0 +1,127 @@
...
+        "encryption": "context",
+        "is not": "secret",
+        "but adds": "useful metadata",
+        "that can help you": "be confident that",
+        "the data you are handling": "is what you think it is",
+    }
+
+    # 4. Generate a 256-bit AES key to use with your keyring.
+    # Here, the input to secrets.token_bytes() = 32 bytes = 256 bits
+    static_key = secrets.token_bytes(32)
```
lucasmcdonald3, line 83:
```suggestion
    # 4. Generate a 256-bit AES key to use with your keyring.
    # In practice, you should get this key from a secure key management system such as an HSM.
    
    # Here, the input to secrets.token_bytes() = 32 bytes = 256 bits
    static_key = secrets.token_bytes(32)
```

examples/src/keyrings/raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,234 @@
...
+    private_key = key.private_bytes(
+        encoding=crypto_serialization.Encoding.PEM,
+        format=crypto_serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=crypto_serialization.NoEncryption()
+    )
+
+    return public_key, private_key
+
+
+def create_rsa_keyring(public_key, private_key):
```
lucasmcdonald3, line 110: I kind of want this to be 

```suggestion
def create_rsa_keyring(public_key_file_name, private_key_file_name):
```

like in https://github.com/aws/aws-encryption-sdk-dafny/blob/mainline/AwsEncryptionSDK/runtimes/net/Examples/Keyring/RawRSAKeyringExample.cs

Users interested in this keyring may bring their own PEM files, and it would be good to meet them where they are.

examples/src/keyrings/raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,234 @@
...
+    for k, v in encryption_context.items():
+        assert v == dec_header.encryption_context[k], \
+            "Encryption context does not match expected values"
+
+    # 8. Demonstrate that the decrypted plaintext is identical to the original plaintext.
+    # (This is an example for demonstration; you do not need to do this in your own code.)
+    assert plaintext_bytes == EXAMPLE_DATA
+
+    # The next part of the example creates a new RSA keyring (for Bob) to demonstrate that
+    # decryption of the original ciphertext is not possible with a different keyring (Bob's)
```
lucasmcdonald3, line 215:
```suggestion
    # decryption of the original ciphertext is not possible with a different keyring (Bob's).
```

examples/src/keyrings/raw_rsa_keyring_example.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # 10. Test decrypt for the original ciphertext using raw_rsa_keyring_bob
+    try:
+        plaintext_bytes_bob, _ = client.decrypt(
+            source=ciphertext,
+            keyring=raw_rsa_keyring_bob
+        )
+
+        raise AssertionError("client.decrypt should throw an error of type CollectionOfErrors!")
+    except CollectionOfErrors:
```
lucasmcdonald3, line 233: Your code is correct, but I don't like that decrypt raises `CollectionOfErrors`.
I would rather have it throw an `AWSEncryptionSDKClientError`, or possibly something more specific.
Could you provide the stack trace for this error?
If you can provide the stack trace, I can make this change in the ESDK.

examples/src/keyrings/raw_rsa_keyring_example.py
RitvikKapila: Makes sense, good catch!

.gitignore
```diff
@@ -31,6 +31,7 @@ __pycache__
 
 # PyTest
 .pytest_cache
+test_keys/
```
lucasmcdonald3, line 34:
```suggestion
# Ignore key materials generated by examples or tests
test_keys/
```

lucasmcdonald3 (commented):
2 nits, fix those then LGTM

.gitignore
lucasmcdonald3: This is so someone looking at the root project's .gitignore has some sense of where this comes from

lucasmcdonald3 (approved):

PR #674, opened 2024-05-03
feat(test_vector_handlers): Hierarchy keyring test vectors

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

Support hierarchy keyrings in test vector manifests.

(Code to generate `decrypt-generate` manifests lives in https://github.com/awslabs/aws-crypto-tools-test-vector-framework/pull/36).

*Testing*

* Generate test vectors for the manifest in Code to generate `decrypt-generate` manifests lives in https://github.com/awslabs/aws-crypto-tools-test-vector-framework/pull/36
* Decrypt generated vectors with keyrings and master keys
* Decrypt generated vectors with .NET ESDK

# Check any applicable:


codebuild/py312/decrypt_hkeyring_with_net.yml
```diff
@@ -0,0 +1,34 @@
...
+      # Download published .NET ESDK so we don't have to build from source
+      - dotnet add NETTestVectors package AWS.Cryptography.EncryptionSDK --version 4.0.1
+
+      # Clone SDK-Dafny repo to get test vectors runner
+      - git clone git@github.com:aws/aws-encryption-sdk-dafny.git
+      - cd aws-encryption-sdk-dafny/AwsEncryptionSDK/runtimes/net/TestVectorsNative/TestVectors
+      - export DAFNY_AWS_ESDK_TEST_VECTOR_MANIFEST_PATH=../../../../../../312_hkeyring_manifest/manifest.json
+  build:
+    commands:
+      - dotnet test --framework net48
```
texastony, line 34: Issue: I thought net48 is only supported on Windows.
@josecorella 
You probably want to use net6.0.

josecorella: net48 is meant to run only on windows. It may succeed on a different OS but it is not guaranteed.

lucasmcdonald3: Yep, this is WIP. I only opened a draft PR so the code is "somewhere" and has "something" running in codebuild CI.

codebuild/py312/decrypt_hkeyring_with_net.yml
```diff
@@ -0,0 +1,55 @@
...
+      # Download previously generated vectors
+      - aws s3 cp s3://generated-vectors-artifacts-bucket/$CODEBUILD_RESOLVED_SOURCE_VERSION/312_hkeyring_manifest.zip 312_hkeyring_manifest.zip
+      - unzip 312_hkeyring_manifest.zip
+      - export DAFNY_AWS_ESDK_TEST_VECTOR_MANIFEST_PATH="${PWD}/312_hkeyring_manifest/manifest.json"
+
+      # Approach 1: "The Right Way":
+
+      # TODO: Get published NET TestVectors runner
+
+      # Approach 2: "The Quick Way:"
```
lucasmcdonald3, line 30:
```suggestion
```

test_vector_handlers/src/awses_test_vectors/manifests/full_message/decrypt.py
```diff
@@ -292,6 +292,7 @@ def from_scenario(
         else:
             master_key_specs = [
                 MasterKeySpec.from_scenario(spec) for spec in raw_master_key_specs
+                if spec["type"] != "aws-kms-hierarchy"
```
josecorella, line 295: ahh i think I answered my own questions and the filtering is done here.

josecorella (changes requested):

codebuild/py312/decrypt_hkeyring_with_masterkey.yml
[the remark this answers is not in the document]
lucasmcdonald3: This uses master keys to decrypt test vectors generated by keyrings.
This really shouldn't be needed, but is good validation.

The manifest has scenarios with the required EC CMM and HKeyrings.
These scenarios aren't compatible with master keys, so the Python test vector runner will filter out test cases with required EC CMM and HKeyrings if using master keys.
There's still lots of scenarios that don't have RECCMM/HKeyring, and master keys are expected to be able to decrypt those.
In theory these scenarios are the same as what we have with other manifests, but it's good coverage to run them anyway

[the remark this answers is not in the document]
josecorella: ahhh gotcha!
the smallest of nits feel free to ignore if it's too much hassle; but shouldn't the file we store in s3 with the hkeyring manifest be called, decrypt_manifest_v4? since the v4 manifest has the reqcmm and hierarchy keyring?

josecorella (approved):

PR #684, opened 2024-05-20
chore(release): update changelog and identifiers.py

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*
chore(release): update changelog and identifiers.py for v3.3.0

# Check any applicable:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,27 @@
 Changelog
 *********
 
+3.3.0 -- 2024-05-20
+===================
+
+Deprecation
+-----------
+The AWS Encryption SDK for Python no longer supports Python 3.7
+as of version 3.3; only Python 3.8+ is supported.
+
+Fixes
+-----------
+* fix: Handle errors when decrypting multiple EDKs with raw RSA MKPs (#672 (https://github.com/aws/aws-encryption-sdk-python/pull/672))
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,27 @@
...
+===================
+
+Deprecation
+-----------
+The AWS Encryption SDK for Python no longer supports Python 3.7
+as of version 3.3; only Python 3.8+ is supported.
+
+Fixes
+-----------
+* fix: Handle errors when decrypting multiple EDKs with raw RSA MKPs (#672 (https://github.com/aws/aws-encryption-sdk-python/pull/672))
```
lucasmcdonald3, line 15: Can you add the commit as well like in previous changelog entries?

```suggestion
* fix: Handle errors when decrypting multiple EDKs with raw RSA MKPs (#672 (https://github.com/aws/aws-encryption-sdk-python/pull/672))  (5f29be9(https://github.com/josecorella/aws-encryption-sdk-python/commit/5f29be92e8e54202a00f13aee6f1c70fb22a6b38))
```

ShubhamChaturvedi7: I couldn't find an automated way to generate the changelog. Given that it's manually written I think adding both the things wouldn't scale and doesn't add any value as well. I chose to add PR link because it has more details and you can anyway refer to the commit from there.

Also if you scroll below you'd see that the format is flexible for the changelog. I'll also check how it was generated the last time - but if it was manual indeed I don't think we should add both.

lucasmcdonald3 (approved):

PR #690, opened 2024-06-18
chore(custom_cmm_example.py): added test for custom_cmm_example.py

RitvikKapila:
added test for v3 default cmm; isort fix for migration examples

*Issue #, if available:*

*Description of changes:*

# Check any applicable:


examples/src/legacy/v3_default_cmm.py
```diff
@@ -0,0 +1,206 @@
...
+)
+from aws_encryption_sdk.key_providers.base import MasterKeyProvider
+from aws_encryption_sdk.materials_managers import DecryptionMaterials, EncryptionMaterials
+from aws_encryption_sdk.materials_managers.base import CryptoMaterialsManager
+
+_LOGGER = logging.getLogger(__name__)
+
+
+@attr.s(hash=False)
+class V3DefaultCryptoMaterialsManager(CryptoMaterialsManager):
```
lucasmcdonald3, line 28: Let's make the class in this file be the SigningOnlyCMM.
The goal is to show customers how to implement a basic custom CMM.

examples/src/legacy/v3_default_cmm.py
```diff
@@ -0,0 +1,206 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
lucasmcdonald3, line 1: Call this `custom_cmm_example.py` or similar

examples/test/legacy/test_i_v3_default_cmm.py
```diff
@@ -0,0 +1,20 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
lucasmcdonald3, line 1: Either put the V3 CMM in this file, or in another file in this folder, then add a test for it in this file.
We don't want to put a copy-paste of the Default CMM in any src/ directory. We're only copy-pasting it for testing, so let's put it in a test directory

examples/src/legacy/v3_default_cmm.py
```diff
@@ -0,0 +1,206 @@
...
+            encryption_context=request.encryption_context,
+        )
+        verification_key = self._load_verification_key_from_encryption_context(
+            algorithm=request.algorithm, encryption_context=request.encryption_context
+        )
+
+        return DecryptionMaterials(data_key=data_key, verification_key=verification_key)
+
+
+def encrypt_decrypt_with_v3_default_cmm(key_arn,
```
lucasmcdonald3, line 161: Let's mirror the method from the Java example, which I think takes in a CMM

examples/test/legacy/v3_default_cmm.py
```diff
@@ -0,0 +1,159 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Copy-paste of the V3 default CMM."""
```
lucasmcdonald3, line 3:
```suggestion
"""Copy-paste of the V3 default CMM with small changes to pass linters."""
```

examples/src/legacy/custom_cmm_example.py
```diff
@@ -0,0 +1,87 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Example to create a custom crypto material manager class."""
```
lucasmcdonald3, line 3:
```suggestion
"""Example to create a custom implementation of the native ESDK CryptoMaterialsManager class."""
```

examples/test/legacy/test_i_custom_cmm_example.py
```diff
@@ -0,0 +1,52 @@
...
+    cmk_arn = get_cmk_arn()
+    botocore_session = botocore.session.Session()
+
+    # Create a KMS master key provider.
+    kms_kwargs = dict(key_ids=[cmk_arn])
+    if botocore_session is not None:
+        kms_kwargs["botocore_session"] = botocore_session
+    master_key_provider = aws_encryption_sdk.StrictAwsKmsMasterKeyProvider(**kms_kwargs)
+
+    # Create the V3 default CMM (V3DefaultCryptoMaterialsManager) using the master_key_provider
```
lucasmcdonald3, line 29: custom signing CMM

tox.ini
```diff
@@ -169,10 +169,11 @@ deps = {[testenv:flake8]deps}
 commands =
     flake8 examples/src/
     flake8 \
-        # Ingore D103 missing docstring errors in tests (test names should be self-documenting)
+        # Ignore D103 missing docstring errors in tests (test names should be self-documenting)
         # E203 is not PEP8 compliant https://github.com/ambv/black#slices
         # W503 is not PEP8 compliant https://github.com/ambv/black#line-breaks--binary-operators
         --ignore D103,E203,W503 \
+        --per-file-ignores 'examples/test/legacy/v3_default_cmm.py: D205,D400,D401' \
```
lucasmcdonald3, line 176: :100:
Maybe add a short comment explaining why this is excluded ("copy-paste, intentionally not changing stuff")

examples/test/legacy/v3_default_cmm.py
RitvikKapila: Good catch. Thanks.

examples/test/legacy/test_i_custom_cmm_example.py
RitvikKapila: Good catch. Thanks!

examples/src/custom_mpl_cmm_example.py
```diff
@@ -0,0 +1,114 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Example to create a custom implementation of the ESDK-MPL ICryptographicMaterialsManager class."""
```
lucasmcdonald3, line 3: Could you expand on this? Unfortunately, this is the only place where we have *any* documentation on writing a custom CMM, so we should put some extra care into this.
I would suggest adding things like
* Custom implementations must implement `get_encryption_materials` and `decrypt_materials`.
* CMMs are composable; if you just want to extend the behavior of the default CMM, you can do this (demonstrated in this example). This is easiest if you just want to add a small check to the CMM methods
* You can also write your own implementation without extending a CMM if your use case calls for it, suggest looking at implementation of default CMM to get a sense of what needs to be done. This is best if you need to fundamentally change aspects of the default CMM.

examples/src/legacy/custom_cmm_example.py
```diff
@@ -0,0 +1,87 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
lucasmcdonald3, line 1: This is all good. You don't need to change this example with my comments on the MPL example.

examples/src/custom_mpl_cmm_example.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Verify that the encryption context used in the decrypt operation includes all key pairs from
+    # the encrypt operation. (The SDK can add pairs, so don't require an exact match.)
+    #
+    # In production, always use a meaningful encryption context. In this sample, we omit the
+    # encryption context (no key pairs).
+    # The encryptor_header.encryption_context has items of the form
+    #     b'key': b'value'
+    # We convert these to strings for easier comparison with the decrypted header below.
+    for k, v in encryptor_header.encryption_context.items():
```
lucasmcdonald3, line 110: You can remove this, this example doesn't have any encryption context.
If it did, you would do this assertion on the encryption context you provide on encrypt, not on the encrypted message's header encryption context.
(Alternatively, you can add some encryption context on encrypt, then assert the decrypted header matches that.)

examples/test/legacy/v3_default_cmm.py
```diff
@@ -0,0 +1,159 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Copy-paste of the V3 default CMM with small changes to pass linters.."""
```
lucasmcdonald3, line 3: (nit)
```suggestion
"""Copy-paste of the V3 default CMM with small changes to pass linters."""
```

examples/src/aws_cryptographic_materials_manager_example.py
```diff
@@ -0,0 +1,133 @@
...
+
+# TODO-MPL: Remove this as part of removing PYTHONPATH hacks.
+MODULE_ROOT_DIR = '/'.join(__file__.split("/")[:-1])
+
+sys.path.append(MODULE_ROOT_DIR)
+
+EXAMPLE_DATA: bytes = b"Hello World"
+
+
+def encrypt_and_decrypt_with_cmm(
```
lucasmcdonald3, line 44:
```suggestion
def encrypt_and_decrypt_with_default_cmm(
```

examples/src/custom_mpl_cmm_example.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Custom CMM implementation using the MPL.
+# This CMM only allows encryption/decryption using signing algorithms.
+# It wraps an underlying CMM implementation and checks its materials
+# to ensure that it is only using signed encryption algorithms.
+class MPLCustomSigningSuiteOnlyCMM(ICryptographicMaterialsManager):
+    """Example custom crypto materials manager class."""
+
+    def __init__(self, keyring: IKeyring) -> None:
```
lucasmcdonald3, line 24: Could this optionally take in a CMM, and if none is provided, then construct a default one with the provided keyring?
Consider
```suggestion
    def __init__(self, keyring: IKeyring, cmm: ICryptographicMaterialsManager) -> None:
        if cmm is None:
             ...
```
The underlying CMM can be *any CMM*. That point comes through clearly by letting this custom CMM take in any CMM

examples/src/aws_cryptographic_materials_manager_example.py
```diff
@@ -0,0 +1,133 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+This example sets up the AWS Cryptographic Material Managers (CMM).
```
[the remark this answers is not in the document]
RitvikKapila: okay

examples/src/custom_mpl_cmm_example.py
RitvikKapila: Thanks

examples/src/custom_mpl_cmm_example.py
RitvikKapila: I am gonna remove the encryption context check, because we're not adding it in other languages.

examples/src/custom_mpl_cmm_example.py
RitvikKapila: Sounds good! Thanks.

examples/src/custom_mpl_cmm_example.py
```diff
@@ -0,0 +1,117 @@
...
+Cryptographic Materials Managers (CMMs) are composable; if you just want to extend the behavior of
+the default CMM, you can do this as demonstrated in this example. This is easy if you just want
+to add a small check to the CMM methods.
+
+Custom implementation of CMMs must implement get_encryption_materials and decrypt_materials.
+If your use case calls for fundamentally change aspects of the default CMM, you can also write
+your own implementation without extending a CMM.
+
+For more information on a default implementation of a CMM,
+please look at the default_cryptographic_materials_manager_example.py example.
```
lucasmcdonald3, line 15:
```suggestion
Cryptographic Materials Managers (CMMs) are composable; if you just want to extend the behavior of
the default CMM, you can do this as demonstrated in this example. This is the easiest approach if you are just adding a small check to the CMM methods, as in this example.

If your use case calls for fundamentally changing aspects of the default CMM, you can also write
your own implementation without extending an existing CMM. The default CMM's implementation is a good
reference to use if you need to write a custom CMM implementation from scratch.

Custom implementations of CMMs must implement get_encryption_materials and decrypt_materials.

For more information on a default implementation of a CMM,
please look at the default_cryptographic_materials_manager_example.py example.
```

Just shuffling some stuff around for clarity

examples/src/custom_mpl_cmm_example.py
```diff
@@ -0,0 +1,117 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+Example to create a custom implementation of the ESDK-MPL ICryptographicMaterialsManager class.
```
lucasmcdonald3, line 4:
```suggestion
Example to create a custom implementation of the MPL's ICryptographicMaterialsManager class
and use it with the ESDK.
```

I like where your head was at here, but it really is just the MPL ICMM.
I think this better communicates that the interface is for the MPL (implicitly, can be used in DBESDK + S3EC once those are out)
and says that this is just using it with ESDK

RitvikKapila: That makes sense. Thanks.

lucasmcdonald3 (approved):

PR #694, opened 2024-06-27
chore(examples): Added raw RSA/AES keyring multithreaded examples

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:


ShubhamChaturvedi7 (approved):
The only shared states I see here are the KeyRing, Client, and Plaintext. So as long as this is what we intended to test, the test execution seems random enough to induce any common threading issues.

src/aws_encryption_sdk/streaming_client.py
```diff
@@ -76,7 +76,8 @@
     # Import internal ESDK modules that depend on the MPL
     from aws_encryption_sdk.materials_managers.mpl.cmm import CryptoMaterialsManagerFromMPL
 
-except ImportError:
+except ImportError as e:
+    print("streamingerror " + str(e))
```
lucasmcdonald3, line 80:
```suggestion
except ImportError:
```

PR #696, opened 2024-08-07
chore: add PyPi API tokens to Cfn template

justplaz:
*Issue #, if available:* n/a

*Description of changes:* updates the CFN template to include the "new" PyPi secrets which move us to API token. this was manually changed in the account for the previous release, but not reflected in CFN. there are some other changes which I am not attempting to fix until we run a release, so as to avoid breaking our release. 

# Check any applicable:


RitvikKapila (approved):
LGTM

Issue #192, opened 2019-09-19
Base64 error messaging in version and type verification


seebees:
When the `version` and `type` elements of the AWS Encryption Message format are verified,
the error message should be customize if the `version` and `type` match the base64 value.
This is convent sugar for a common problem.

Valid blobs will have the first 2 bytes in HEX will be: `01`, `80`, and in DEC `01`, `128`
Valid blobs that have not been properly base64 decoded
the values in HEX will be `41`, `59`, and in DEC `65`, `89`

https://github.com/awslabs/aws-encryption-sdk-javascript/pull/217

PR #697, opened 2024-08-12
fix(Examples): Validate EC on decrypt

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:

### DIFF
--- a/examples/src/aws_kms_discovery_keyring_example.py
+++ b/examples/src/aws_kms_discovery_keyring_example.py
@@ -161,6 +161,11 @@ def encrypt_and_decrypt_with_keyring(
         encryption_context=encryption_context,
     )
 
+    # 9. Demonstrate that the decrypted plaintext is identical to the original plaintext.
+    # (This is an example for demonstration; you do not need to do this in your own code.)
+    assert plaintext_bytes == EXAMPLE_DATA, \
+        "Decrypted plaintext should be identical to the original plaintext. Invalid decryption"
+
     # 11. Demonstrate that if a discovery keyring (Bob's) doesn't have the correct AWS Account ID's,
     # the decrypt will fail with an error message
     # Note that this assumes Account ID used here ('888888888888') is different than the one used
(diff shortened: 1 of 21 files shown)


examples/src/aws_kms_discovery_keyring_example.py
```diff
@@ -153,18 +153,15 @@ def encrypt_and_decrypt_with_keyring(
     # successfully decrypted. The resulting data key is used to decrypt the
     # ciphertext's message.
     # If all calls to KMS fail, the decryption fails.
-    plaintext_bytes, dec_header = client.decrypt(
+    plaintext_bytes, _ = client.decrypt(
         source=ciphertext,
-        keyring=discovery_keyring
+        keyring=discovery_keyring,
+        # Verify that the encryption context in the result contains the
+        # encryption context supplied to the encrypt method
```
RitvikKapila, line 160: According to Ryan, we want to tell the customer to always provide the encryption context during decrypt. I wonder if it'll be better if we remove the word "Verify" and just say "Enter the encryption context supplied to the encrypt method".

Same goes for Line #192 in this file and all other examples.

RitvikKapila (commented):
This looks good.
One thing, do we not want to make the changes of providing the encryption context during decrypt in other migration examples? For instance:
- migration_raw_rsa_key_example.py
- migration_raw_aes_key_example.py
- migration_aws_kms_key_example.py

I think we should, to point out explicitly that they should provide the encryption context while calling decrypt.

examples/src/aws_kms_discovery_keyring_example.py
lucasmcdonald3: Updated, I slightly preferred "Provide" instead of "Enter"

Good catch on migration examples, I updated those to supply EC on decrypt if decrypting with keyring.
I also added manual EC validation if decrypting with MKP since MKPs don't support EC validation on decrypt.

RitvikKapila (approved):
very small nit: I would change the name of the encrypted header, e.g., `enc_header_mkp` to `encrypted_header_mkp` just because we have the decrypted header variable as `decrypted_header_keyring_using_mkp`. However, it's not a big thing if we keep this as is.

Otherwise this LGTM. Approved.

PR #785, opened 2026-01-09
chore(CI): fix CI

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*

- Issue 1: Retired `macos-13`:
	- Used macos-latest
- Issue 2: Python <3.11 incompatible with macOS latest runner
	- Excluded macos latest when python<3.11 
	- TODO: use different runner when on python 3.11
- Issue 3: pytest using outdated Python ast module until pytest 7.3.1
	- This has been fixed on pytest>7.3.1 by https://github.com/pytest-dev/pytest/issues/10977
	- Moved out pytest deps to use 7.4.0 instead of 7.2.1 but see issue 4.
- Issue 4: Warning about pytest.warns(None) being deprecated on 7.4.0. In pytest 8+, passing None to pytest.warns() will be completely removed.
	- So, decided to move pytest deps to use 8.0.0 and migrate `pytest.warns(None)`
	- Note that I did not move the deps to use 9.x which is latest as it requires Python >=3.10 but we still support Python < 3.10 and have to test against it.
- Issue 5: `-rc` flag removed in isort command
	- rc flag is not required anymore, recursive is used by default without the flag.
	- See https://github.com/PyCQA/isort/issues/1263
- Issue 6: Memory error in 32 bit windows machine.
	- Test `test_GIVEN_two_distinct_valid_encrypted_data_keys_WHEN_serialize_encrypted_data_keys_THEN_deserialize_equals_inputs` errored out because of memory error in 32 bit windows OS. 
	- Separated these tests into batches which also reduces duplicate parameter in the test.
- Issue 7: Timeout in test vector:
	- We timeout "awses_local.yml" workflow in 2 hours which was still not enough for some python version (example: 3.8). 
	- Increased it to 2.5 hours. 
	- Added following to the handoff doc under "Quality of life CI improvement"
		- research to decrease time to run "awses_local.yml" workflow which currently takes more then 2 hour.
- Issue 8 (Flaky not busted): CodeBuild rate limit errors
	- Add concurrency control to cancel previous workflow runs on new commits, preventing CodeBuild rate limit errors from too many simultaneous builds.

- Removed redundant run of `ci_codebuild-tests.yml`. This was getting run with `pull.yml` and `on pull`.
- Used black for formatting

# Check any applicable:

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -4,6 +4,9 @@ name: Daily CI
 on:
   schedule:
     - cron: "00 15 * * 1-5"
+  pull_request:
+    paths:
+      .github/workflows/daily_ci.yml
 
 permissions:
   contents: read


.github/workflows/ci_tests.yaml
```diff
@@ -47,11 +47,23 @@ jobs:
...
+          # https://github.com/actions/setup-python/issues/948
+          - python: 3.7
+            os: macos-latest
+          - python: 3.8
+            os: macos-latest
+          - python: 3.9
+            os: macos-latest
+          - python: 3.10
+            os: macos-latest
+
```
ShubhamChaturvedi7: We should find a way to run these before we merge it.

Issue #793, opened 2026-03-25
_set_signature_type returning "Unsupported signing algorithm info" due to isinstance check


jvan1997:
### Security issue notifications
If you discover a potential security issue in the AWS Encryption SDK we ask that you notify AWS Security via our [vulnerability reporting page](https://aws.amazon.com/security/vulnerability-reporting/). Please do **not** create a public GitHub issue.

### Problem:

`aws-encryption-sdk` version 4.x introduced a regression in `_set_signature_type()` via [PR #467](https://github.com/aws/aws-encryption-sdk-python/pull/467/commits/c0eb6b769198c6b4b6eec62f34ea90c3626212d2) ([specific line](https://github.com/aws/aws-encryption-sdk-python/pull/467/commits/c0eb6b769198c6b4b6eec62f34ea90c3626212d2#diff-ae2f425c065b9fd27ed1631b058ecd67a16fb131e445977719e6cb1e3396cf7eR49)) that replaced a `verify_interface` call with a bare `isinstance` check:

```python
def _set_signature_type(self):
    if not isinstance(self.algorithm.signing_algorithm_info, type(ec.EllipticCurve)):
        raise NotSupportedError("Unsupported signing algorithm info")
    return ec.EllipticCurve
```

`type(ec.EllipticCurve)` evaluates to `ABCMeta`. This worked when `cryptography` implemented EC curve classes as pure Python ABCs, where `isinstance(ec.SECP384R1, ABCMeta)` returned `True` because the class itself is an instance of its metaclass.

However, in newer versions of `cryptography`, `ec.SECP384R1` is backed by Rust internals via `rust_openssl`, and its metaclass is `type`, not `ABCMeta`. This causes both of the following to return `False` at runtime:

```
isinstance(ec.SECP384R1, type(ec.EllipticCurve))  # False — metaclass is 'type', not ABCMeta
isinstance(ec.SECP384R1, ec.EllipticCurve)         # False — it's a class, not an instance
```

This raises `NotSupportedError("Unsupported signing algorithm info")` when decrypting with `AES_256_GCM_HKDF_SHA512_COMMIT_KEY_ECDSA_P384`, which is used by clients such as the Odin encryption client.

[...]

PR #800, opened 2026-04-22
fix: use issubclass instead of isinstance in _set_signature_type

texastony:
## Summary

Replace `isinstance(signing_algorithm_info, type(ec.EllipticCurve))` with `issubclass(signing_algorithm_info, ec.EllipticCurve)` wrapped in `try/except TypeError` in `_PrehashingAuthenticator._set_signature_type`.

The old check used `isinstance` against `ABCMeta` which was fragile and semantically incorrect for checking class hierarchy.

## Changes

- **authentication.py**: Fix `_set_signature_type` to use `issubclass` with `TypeError` guard
- **test_crypto_prehashing_authenticator.py**: Use real EC curve classes instead of MagicMock; add `test_set_signature_type_elliptic_curve_known_value`
- **test_crypto_authentication_signer.py**: Update mock patterns to use real `ec.SECP256R1` for `signing_algorithm_info`
- **test_crypto_authentication_verifier.py**: Same mock pattern fix

## Testing

All 4228 unit tests pass.

Based on #794, extended to also fix signer/verifier tests that used the same broken mock pattern.

Closes #794 ; stages #793 for release.

### DIFF
--- a/src/aws_encryption_sdk/internal/crypto/authentication.py
+++ b/src/aws_encryption_sdk/internal/crypto/authentication.py
@@ -36,7 +36,10 @@ class _PrehashingAuthenticator(object):
 
     def _set_signature_type(self):
         """Ensures that the algorithm signature type is a known type and sets a reference value."""
-        if not isinstance(self.algorithm.signing_algorithm_info, type(ec.EllipticCurve)):
+        try:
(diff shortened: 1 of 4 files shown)


sharmabikram (approved):
Looks good to me

PR #817, opened 2026-09-08
feat: add PEP 561 py.typed marker to support mypy type analysis (#563)

### LINKED ISSUE #563: Feature request: Add support typing / mypy
### Problem:

Using the aws_encryption_sdk in a project with mypy produces this error:

```
error: Skipping analyzing "aws_encryption_sdk": module is installed, but missing library stubs or py.typed marker
```

### Solution:

I attempted to simply add the py.typed marker file to my virtualenv and it seems not as simple, a lot of other incompatibilities.
Maybe stubs or .pyi could be provided

Thank you very much!

ShyamD2:
*Issue #, if available:*
Fixes #563
*Description of changes:*
Adds PEP 561 compliance so downstream projects using `mypy` can analyze `aws_encryption_sdk` without receiving `error: Skipping analyzing "aws_encryption_sdk": module is installed, but missing library stubs or py.typed marker`.

### Summary of Changes:
- Added `src/aws_encryption_sdk/py.typed` containing the `partial` marker per PEP 561.
- Updated `setup.py` to include `package_data={"aws_encryption_sdk": ["py.typed"]}` so wheel packages bundle the marker.
- Updated `MANIFEST.in` with `include src/aws_encryption_sdk/py.typed` for source distribution packaging.
- 
### Verification:
Tested with `mypy` analyzing public client initialization:
```bash
$ mypy -c "import aws_encryption_sdk; client = aws_encryption_sdk.EncryptionSDKClient()"
Success: no issues found in 1 source file

### CODE REVIEW

**Comment by kessplas** [comment]
Hello @ShyamD2 ,

Thanks for your contribution. As you point out, the ESDK for Python does not currently comply with PEP 561 or support Mypy in any meaningful way. However, we do not have any immediate plans to add new type hints, so it doesn't make sense to add these files to advertise typing information which is not actually present. 

If you want to avoid the error, you can add this to your local configuration: 

```
[[tool.mypy.overrides]]
module = "aws_encryption_sdk.*"
ignore_missing_imports = true
```

Alternatively, you propose in https://github.com/aws/aws-encryption-sdk-python/issues/563 to add .pyi stubs for the primary public API. This would be a more reasonable change. If you'd prefer this approach, please add .pyi stub files accordingly.

ShyamD2:
Hello @kessplas,

Thanks for the feedback and guidance! That makes total sense—advertising typing via py.typed without actual type definitions doesn't provide real type safety downstream.

Per your suggestion, I have updated this PR with .pyi stub files covering the primary public API:

aws_encryption_sdk/__init__.pyi: Type definitions for EncryptionSDKClient, EncryptionSDKClientConfig, methods (encrypt, decrypt, and @overload definitions for stream), along with top-level re-exports.
identifiers.pyi & structures.pyi: Stubs for AlgorithmSuite / Algorithm, CommitmentPolicy, and MessageHeader.
streaming_client.pyi: Stubs for StreamEncryptor, StreamDecryptor, and config classes.
materials_managers/*.pyi & exceptions.pyi: Supporting public interfaces.
Updated setup.py and MANIFEST.in to ensure .pyi stubs and py.typed are bundled in both source distributions and wheels.
Verified with mypy resolving types correctly for client initialization, encryption, decryption, and streaming flows. Looking forward to your review!

